PR #13, opened 2018-08-24
use individual flags

petrelharp:


bhaller:
Do we want the flags in treerec/tskit/slim.h?  Would that file get pushed back into tskit, so that the flags we use are a public part of tskit's codebase?  If not, perhaps slim_global.h would be a more appropriate place for them?

core/slim_sim.cpp
```diff
@@ -4939,6 +4952,8 @@ void SLiMSim::AddIndividualsToTable(Individual * const *p_individual, size_t p_n
             memcpy(p_tables->individuals->metadata
                     + p_tables->individuals->metadata_offset[msp_individual],
                     &metadata_rec, sizeof(IndividualMetadataRec));
+            uint32_t current_flags = p_tables->individuals->flags[msp_individual];
+            p_tables->individuals->flags[msp_individual] = (current_flags | flags);
```
bhaller: In C these two lines would more conventionally be written as:

p_tables->individuals->flags[msp_individual] |= flags;

unless current_flags is used again below, but I don't see that it is.  Easier to see that all that is really happening is that flags is getting OR'd into the existing flags.

Issue #26, opened 2018-11-11
Add support for installation in Bioconda


bhaller:
Apparently Bioconda (https://bioconda.github.io) is a thing people often use to install software in the Linux and Mac worlds.  A user has requested that we add SLiM to it.  This would require more Linux experience, etc., than I really have, I think, but @petrelharp has expressed possible willingness to take it on.  I'm not sure how this normally works, but I imagine we would only want tagged releases of SLiM to be available through Bioconda; perhaps each time we did a new tagged release we would need to submit a new Bioconda configuration file or something...?

molpopgen:
I just did a quick hack of a recipe, but there is a road block.  It looks like the cmake system is not set up to generate install targets:

```
cmake -D CMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME/tmp ../SLiM-3.2
make -j 6 > /dev/null
make install
```

The output is:
```
make: *** No rule to make target 'install'.  Stop.
```

That needs to be fixed before a working recipe can be generated.

bhaller:
@jgallowa07, you set up the cmake stuff with Boyana, right?  Any chance of you guys doing a pull request to provide a "make install" target for it?

@molpopgen, thanks for looking into this!  Are you offering to manage the Bioconda submissions for SLiM?  That would of course be greatly appreciated!  :->  We do a release every couple of months, so it wouldn't be a big thing, but I personally find a lot of this very Unix-y stuff a bit mystifying.  I'm used to working in my Xcode IDE.  :->

molpopgen:
@bhaller -- Happy to do a first deposit to bioconda.  Not sure I can keep up with later releases.  I'm already maintaining several recipes, and @jeromekelleher and I have to deal with an issue that is preventing msprime from being used as a dependency right now (annoying!!), which will likely affect the chances of getting the slim Python package into bioconda.

You will see that the recipes are rather simple.  libsequence is similar to slim in that it has no external dependencies beyond a compiler.  The files are [here](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/libsequence), and I basically just hacked those to make one for slim, but ran into the "make install" issue, which is why there's no PR.

The only real difference for slim is to change `build.sh` to use cmake and to remove skipping the building on os x.

petrelharp:
I've added an `install` target to cmake, and updated the README that explains it... does that give you enough to go on, @molpopgen?

molpopgen:
[...]

PR #27, opened 2018-11-19
added install cmake target

petrelharp:
This is to allow #26 to move forward. Perhaps folks could test this on their own platforms?  (See the README for instructions.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,20 +30,19 @@ https://gitlab.kitware.com/cmake/community/wikis/FAQ#out-of-source-build-trees
 
 CMake is available with `port install cmake` or `brew install cmake` (on OS X, with MacPorts or Homebrew respectively), or with `aptitude install cmake` (on debian).
 
-To compile and install the development version from this repository, do:
+To compile the development version from this repository, do:
 
 	git clone https://github.com/MesserLab/SLiM.git
 	mkdir SLiM_build
 	cd SLiM_build
 	cmake -D CMAKE_BUILD_TYPE=Release ../SLiM
 	make
-    make install
 
-The `make install` step will install the `slim` and `eidos` binaries to appropriate system directories (hopefully).
-If you don't want this to happen, just skip this step; the binaries
-will be located within the `SLiM_build` directory.
+The resulting binaries (`slim` and `eidos`) will be located in the `SLiM_build` directory.
 You can name the build directory anything you wish; there is nothing magical about the name `SLiM_build`.
-To install the binaries to a different directory (like the `--user` flag), instead run
+To install these to the default system directories, you can just run `make install`;
+but this will probably fail because of permissions. Instead you should tell cmake where to install the binaries
+(analogous to the `--user` flag):
 
 	cmake -D CMAKE_BUILD_TYPE=Release -D -DCMAKE_INSTALL_PREFIX:PATH=/path/to/install ../SLiM
 	make
--- a/README.md
+++ b/README.md
@@ -40,15 +40,16 @@ To compile the development version from this repository, do:
 
 The resulting binaries (`slim` and `eidos`) will be located in the `SLiM_build` directory.
 You can name the build directory anything you wish; there is nothing magical about the name `SLiM_build`.
-To install these to the default system directories, you can just run `make install`;
-but this will probably fail because of permissions. Instead you should tell cmake where to install the binaries
-(analogous to the `--user` flag):
+Running `make install` will then attempt to install these to the default system directories,
+but this will probably fail unless you run it with root permissions (e.g., by doing `sudo make install`). 
+To explicitly tell cmake where to install the binaries, run:
 
 	cmake -D CMAKE_BUILD_TYPE=Release -D -DCMAKE_INSTALL_PREFIX:PATH=/path/to/install ../SLiM
 	make
     make install
 
-where `/path/to/install` is the path where you want `slim` and `eidos` installed below (they would be put in `/path/to/install/bin` in this example).
+where `/path/to/install` is the path where you want `slim` and `eidos` installed below 
+(they would be put in `/path/to/install/bin` in this example, analogous to the `--user` flag).
 
 To update the build, in the `SLiM_build` directory run:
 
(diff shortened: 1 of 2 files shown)


bhaller:
Looking at the diffs here, it looks like cmake installs, by default, to a location determined by a magic DESTINATION variable that is (I would hope) correct on the local platform?  So in the docs, I guess we can't say "make install puts the built executables at X", because X is platform-dependent; we would instead say to do `which slim` to find out where it was put after it was built?  Or what?  And is there any guarantee that the location where the install goes is in the user's path (if they haven't screwed up the default path, at least)?  Sorting out path issues can be confusing for novice Unix users (like, ahem, me).  If the installed executable is not in the user's path, then `which slim` won't even find it; how do they then find out where it was put?  Maybe that's a dumb question, but I have been puzzled by this sort of thing more than once.  :->

bhaller:
Also, I just tried running this on OS X, and got this error:

```
darwin-447:~/Desktop/SLiM_build bhaller $ make install
[ 47%] Built target gsl
[ 57%] Built target tables
[ 70%] Built target eidos
[ 99%] Built target slim
[100%] Built target kastore
Install the project...
-- Install configuration: "Release"
-- Installing: /usr/local/bin/slim
CMake Error at cmake_install.cmake:36 (file):
  file INSTALL cannot copy file "/Users/bhaller/Desktop/SLiM_build/slim" to
  "/usr/local/bin/slim".

make: *** [install] Error 1
```

I believe the issue is permissions; /usr/local/bin/ is the right place to install to on OS X (that's where the SLiM installer puts it, too), but it requires root permissions to write to it, so the user will typically need to do `sudo make install` for the command to succeed (and that does, indeed, succeed for me).  So... should the instructions here recommend using `sudo`, to avoid confusion?

It looks like cmake does print out where it is installing things to, so that answers my question above about how the user can figure that out.  Of course the path issue may still exist; probably the docs here should mention that, so that clueless users can at least have a pointer as to what to do if the installed build is not in their path...?

molpopgen:
> Looking at the diffs here, it looks like cmake installs, by default, to a location determined by a magic DESTINATION variable that is (I would hope) correct on the local platform? So in the docs, I guess we can't say "make install puts the built executables at X", because X is platform-dependent; we would instead say to do `which slim` to find out where it was put after it was built? Or what? And is there any guarantee that the location where the install goes is in the user's path (if they haven't screwed up the default path, at least)? Sorting out path issues can be confusing for novice Unix users (like, ahem, me). If the installed executable is not in the user's path, then `which slim` won't even find it; how do they then find out where it was put? Maybe that's a dumb question, but I have been puzzled by this sort of thing more than once. :->

The location is handled at the execution of cmake as in my [comment](https://github.com/MesserLab/SLiM/issues/26#issuecomment-439638719) in #26.  The default is the usual place for the system, which typically would require permissions.  A user is responsible for prepending to PATH if they put things somewhere "creative".

bhaller:
OK, that all sounds good.  Re:

> which typically would require permissions

So, shall we tell users to do "sudo make install" rather than "make install", then?  That seems best to me, to avoid confusing novice users.  @petrelharp @molpopgen agreed?

molpopgen:
> So, shall we tell users to do "sudo make install" rather than "make install", then? That seems best to me, to avoid confusing novice users. @petrelharp @molpopgen agreed?

No, it depends where they wish to install.  Installing as sudo into ~/bin would be a mistake, for example.

bhaller:
> No, it depends where they wish to install. Installing as sudo into ~/bin would be a mistake, for example.

Well, it wouldn't do any harm, would it?  It would just be unnecessary.  I imagine users who are installing into custom locations know what they're doing anyway; it's the users installing into the default location (which typically requires sudo) who are likely to be clueless and need hand-holding (and thus should be told to do `sudo make install`).  I don't like the idea of telling users to type a series of commands that will probably fail on their machine; that will just lead to annoying support emails and such.

bhaller:
> I'm going to guess that a big application of this type of install will be on HPC systems. In those cases, the `sudo` will fail.

Interesting.  And it would succeed without the `sudo`?  Or it fails either way, and they need to ask the system administrator to install it for them?

molpopgen:
sudo fails there because users are not given access to it, so the only
workable method is to install wherever they are allowed, or wait for IT to
install it.  Same goes for a student working on the lab's workstation, etc.

bhaller:
> sudo fails there because users are not given access to it, so the only workable method is to install wherever they are allowed, or wait for IT to install it. Same goes for a student working on the lab's workstation, etc.

OK; so then telling them to try to use `sudo` won't really do any harm; they wouldn't have been able to install it without `sudo` either.  If anything, the fact that they won't be able to use `sudo`, and will get an error message telling them so, will tell them that they need to find an administrator to install the software for them.  No?  I'm still not seeing the case for not telling people to use `sudo`.

README.md
```diff
@@ -30,15 +30,26 @@ https://gitlab.kitware.com/cmake/community/wikis/FAQ#out-of-source-build-trees
...
+    make install
 
-The resulting binary will be called `slim`, within the `SLiM_build` directory.  You can name the build directory anything you wish; there is nothing magical about the name `SLiM_build`.
+The `make install` step will install the `slim` and `eidos` binaries to appropriate system directories (hopefully).
+If you don't want this to happen, just skip this step; the binaries
+will be located within the `SLiM_build` directory.
+You can name the build directory anything you wish; there is nothing magical about the name `SLiM_build`.
+To install the binaries to a different directory (like the `--user` flag), instead run
+
+	cmake -D CMAKE_BUILD_TYPE=Release -D -DCMAKE_INSTALL_PREFIX:PATH=/path/to/install ../SLiM
```
petrelharp: here's how to tell it where you want to install

petrelharp:
I've rewritten the README to more explicitly address this tripping point - does it work for you now, Ben?

README.md
bhaller: How about "but this will probably fail because of permissions unless you run it with root permissions (i.e., `sudo make install`).  You can instead tell cmake..."

I think we ought to explicitly point people toward using sudo.  On an HPC cluster or some such, they may not have the ability to do that (but might want to ask an admin to do it for them), but on a user's local machine that is what they will probably want to do, so we should make the procedure explicit.

petrelharp: I've adjusted this. How is it now?

bhaller: Looks good, thanks.  I'll accept these pull requests eventually; I'm not waiting for anything in particular, but I'm busy being on vacation, and there's no urgency to them as far as I'm aware of them, so I'll deal with them later.  If anybody is impatient for them to go in sooner, let me know.  :->

PR #36, opened 2019-03-19
Add nucleotides

petrelharp:
This is space to discuss additional things related to nucleotide support.


petrelharp:
Things to do related to tree sequences plus nucleotides:

- update pyslim
- deal with input (output should be ok already)
- update text output/input
- write ancestral sequence to kastore reference sequence (to table collection dump) as [implemented here](https://github.com/MesserLab/SLiM/pull/36/commits/19d628933572df5187a1c9d68ac3916142e1273c#diff-164c8a503989aafdda54a6ad5ffc8b9fR6338)

PR #39, opened 2019-04-30
tskit update

petrelharp:
This is not done yet, but it (a) successfully built after moving over kastore, and (b) I seem to be on my way to porting over to the new tskit API. There's a lot of changes, and we'll need to do a good check for changes in memory management even after I get everything to compile.


bhaller:
Wow, these changes are much more extensive than I realized they would be.  Thanks for dealing with this, Peter.  Indeed, once this diff is finalized I'll run it all through my test suite, and I'll also do some leaks analysis.

bhaller:
The only change I would suggest thus far is for you to add a line to the VERSIONS file, at the bottom of the list titled "development head", saying something like "merging in tskit and kastore version XXXXX", so there's a record of what version was merged in and when.  Thanks!

petrelharp:
ok, @bhaller - this builds! I have not tested anything, except that it runs.  Could you run some checks?

The main thing that I think we need to worry about having changed is memory management: for instance, previously we had to `table_collection_alloc( )` and then `table_collection_copy( )`; now by default `tsk_table_collection_copy( )` will also alloc things (although this is called `tsk_individual_table_init( )` now), unless you ask it not to (with the `TSK_NO_INIT` flag).  Hopefully I have got this all right, but...

bhaller:
So, I want to test this, but at the moment I’m flailing around trying to figure out how to handle it in git.  Can you get onto Slack for a sec?

petrelharp:
Minor question here for @jeromekelleher: tskit includes its own header files like `include <tskit/trees.h>`; but doing things that way, instead of including with `" "`, why aren't we worried that if a user has a different `tskit` installed somewhere in the default Include paths, that they won't end up with the wrong versions?

bhaller:
OK @petrelharp, I've just done four commits.  One has fixes for warnings, which should probably get pushed back into tskit (but Jerome may want to fix them in a different way).  One fixes a leak; not sure whether it was pre-existing or introduced by these tskit changes, but anyway it's fixed now.  :->  One gets the Xcode project working again, and one just has a few lines of debugging cruft that are commented out.

At present I am unable to do any testing involving pyslim, because of the "No module named 'kastore'" error I'm getting when trying to `import pyslim`.

The remaining issue that I did see is that saving .trees to text and then loading it no longer works.  It looks like an "id" column has been added to some/all of tskit's `dump_text` functions, such as `tsk_edge_table_dump_text()`.  The corresponding text load functions, like `edge_table_load_text()`, are not expecting that id column to be there, and so they error out because the header doesn't match the expected header.  I'm not sure whether we need to actually do something with those id columns, or whether we can just read them in and discard the values we see.  If the latter, I can probably do that myself, if you like; but if the id column actually had some significance and needs to be handled, then maybe that's in your court.

Apart from that issue, this looks good so far.  The text issue and the kastore issue are, together, preventing me from running many tests, though, so let's work through them and then I'll do another testing pass.

petrelharp:
>  I'm not sure whether we need to actually do something with those id columns, or whether we can just read them in and discard the values we see.

We don't need those columns.

bhaller:
> > I'm not sure whether we need to actually do something with those id columns, or whether we can just read them in and discard the values we see.
> 
> We don't need those columns.

OK, I'll fix the text-reading stuff to ignore that column, then, unless you intend to do it.  I figure I will not attempt to preserve backward compatibility (i.e., the ability to read the tables in the old format too); text format is for testing, not for production.  Agreed?

petrelharp:
I've added the `id` column to the edge table text reading code; seems to work, now?

jeromekelleher:
> Minor question here for @jeromekelleher: tskit includes its own header files like include <tskit/trees.h>; but doing things that way, instead of including with " ", why aren't we worried that if a user has a different tskit installed somewhere in the default Include paths, that they won't end up with the wrong versions?

This is a good point, and tskit probably should use quotes when including the internal header files.

VERSIONS
```diff
@@ -84,6 +84,7 @@ development head:
 	correct recipe 16.15, which did not match the manual
 	optimize operator subset [] with a singleton integer argument, since that is the common case
 	fix an ordering problem with the remembered genomes after loading a .trees file (didn't match the order in the individuals table)
+    merge in kastore 67c5a6af4b85ea263e5cab854d94b691949e88ae and tskit 3d16a36f8e54ac01a86f182dc73103bd07badf3e
```
jeromekelleher: Do these git hashes correspond to the current tagged releases? Version numbers would be more useful than hashes I think.

jeromekelleher:
> One has fixes for warnings, which should probably get pushed back into tskit (but Jerome may want to fix them in a different way).

Would you mind opening an issue on these warnings please @bhaller?

VERSIONS
bhaller: agreed

bhaller:
> > One has fixes for warnings, which should probably get pushed back into tskit (but Jerome may want to fix them in a different way).
> 
> Would you mind opening an issue on these warnings please @bhaller?

OK, done: https://github.com/tskit-dev/tskit/issues/182

I think only one is of interest to you, actually; the rest are in SLiM's code, I think.

bhaller:
> > Minor question here for @jeromekelleher: tskit includes its own header files like include <tskit/trees.h>; but doing things that way, instead of including with " ", why aren't we worried that if a user has a different tskit installed somewhere in the default Include paths, that they won't end up with the wrong versions?
> 
> This is a good point, and tskit probably should use quotes when including the internal header files.

The GSL source code uses angle brackets for including its own headers; I think there might be a reason for doing this in library code.  Certainly when a GSL header A includes another GSL header B it _has_ to use angle brackets, otherwise when client code includes header A the include of header B would fail, if it used quotes, right?  Because header B would not be on the non-system header search path.  Given that, you might as well use angle brackets everywhere, for consistency; at least then the tskit headers only need to be findable with <>, not with _both_.  We figured out how to get the header search path sorted out so that the angle-bracket includes worked properly in SLiM.  Here's a useful article on the topic:

https://commandlinefanatic.com/cgi-bin/showarticle.cgi?article=art026

It sounds like this header search path stuff is not explicitly specified/guaranteed by the standard, but GCC's behavior is clear and Clang follows it.  That is good enough for SLiM, but maybe not for tskit; but on the other hand I think you really don't have any choice but to use angle brackets, because of the issue I mentioned above.  So I recommend no change.

jeromekelleher:
> It sounds like this header search path stuff is not explicitly specified/guaranteed by the standard, but GCC's behavior is clear and Clang follows it. That is good enough for SLiM, but maybe not for tskit; but on the other hand I think you really don't have any choice but to use angle brackets, because of the issue I mentioned above. So I recommend no change.

I think I looked this up at the time and came to the same conclusion. Thanks for the summary, very helpful.

bhaller:
> I've added the `id` column to the edge table text reading code; seems to work, now?

Yes, works for me as well.  I guess only the edge table changed; I thought this id column had been added to other tables too, but maybe they already had it.  :->

I just found and fixed another leak, in the text code, so that was worth checking.  I'm also seeing some leaks unrelated to tree-seq recording; it has been a long time since I checked for leaks, that was a good idea.  I'll fix them over in the main branch, though.

And with that, I think this pull request is just about ready to go.  The only remaining nit is that in the VERSIONS file, as Jerome pointed out, it would be more helpful to have version numbers for tskit and kastore, if possible?  Or both?  Anyway, as soon as you give a final thumbs-up, @petrelharp, I'll merge it in.

VERSIONS
petrelharp: They *don't*, because I was trying to get the latest bugfixes. For instance, that tskit hash is
```
commit 3d16a36f8e54ac01a86f182dc73103bd07badf3e (HEAD -> master, origin/master, origin/HEAD)
Merge: 39f3270 87c30c3
Author: Jerome Kelleher <jk@well.ox.ac.uk>
Date:   Fri Apr 26 09:13:27 2019 -0400

    Merge pull request #180 from jeromekelleher/mem-issues
    
    Fix heap overflow in deduplicate sites.
```

... what do you think?

The `kastore` hash is tagged `C_1.1.0`, however.

bhaller: OK, I'll revise the VERSIONS file to say:

> merge in kastore C_1.1.0 (67c5a6af4b85ea263e5cab854d94b691949e88ae) and tskit post-0.1.5 (3d16a36f8e54ac01a86f182dc73103bd07badf3e)

Just nice to have a more concrete reference to what we've got.  :->

PR #42, opened 2019-06-26
enable pedigree id access when treeseq is enabled

petrelharp:
When treeseq is enabled, we have pedigree IDs, so that every individual has a unique ID. These we get to see in the output, in the individual metadata. However, they're not available in Eidos because at the various places it might be obtained there's a check to see if pedigree recording is enabled. (e.g. [here](https://github.com/MesserLab/SLiM/blob/6d543b6337398996de4d95181128a9a52117175e/core/individual.cpp#L220)).  This allows access when tree sequence recording is enabled also, which is important for being able to connect the individuals in SLiM to the individuals in the tree sequence. The docs would need to be updated also.

### DIFF
--- a/core/genome.cpp
+++ b/core/genome.cpp
@@ -549,7 +549,7 @@ EidosValue_SP Genome::GetProperty(EidosGlobalStringID p_property_id)
 		case gID_genomePedigreeID:		// ACCELERATED
 		{
 			if (!subpop_->population_.sim_.PedigreesEnabledByUser())
-				EIDOS_TERMINATION << "ERROR (Genome::GetProperty): property genomePedigreeID is not available because pedigree recording has not been enabled." << EidosTerminate();
+				EIDOS_TERMINATION << "ERROR (Genome::GetProperty): tree sequence recording and/or pedigree recording must be enabled for property genomePedigreeID to be available." << EidosTerminate();
 			
 			return EidosValue_SP(new (gEidosValuePool->AllocateChunk()) EidosValue_Int_singleton(genome_id_));
 		}
@@ -612,8 +612,9 @@ EidosValue *Genome::GetProperty_Accelerated_genomePedigreeID(EidosObjectElement 
 	{
 		Genome *value = (Genome *)(p_values[value_index]);
 		
-		if (!value->subpop_->population_.sim_.PedigreesEnabledByUser())
-			EIDOS_TERMINATION << "ERROR (Genome::GetProperty): property genomePedigreeID is not available because pedigree recording has not been enabled." << EidosTerminate();
+		if (!(value->subpop_->population_.sim_.PedigreesEnabledByUser()
+              || value->subpop_->population_.sim_.RecordingTreeSequence()))
+			EIDOS_TERMINATION << "ERROR (Genome::GetProperty): tree sequence recording and/or pedigree recording must be enabled for property genomePedigreeID to be available." << EidosTerminate();
 		
 		int_result->set_int_no_check(value->genome_id_, value_index);
 		++value_index;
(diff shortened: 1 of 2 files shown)


bhaller:
Hmm.  If you want them, just enable pedigree tracking, no?  I didn't want to formally enable pedigree tracking as a side effect of enabling tree-sequence recording, because that might entail additional overhead in future (if pedigree tracking becomes more complicated in future).  If you want the feature, you can/should enable the feature.

petrelharp:
>  I didn't want to formally enable pedigree tracking as a side effect of enabling tree-sequence recording, because that might entail additional overhead in future.

That makes sense to me, and I don't think we should formally enable pedigree tracking for tree sequence recording.

But: am I right, that currently there is no way to definitively connect the individuals in SLiM to the individuals in the tree sequence? Like, say that there's information in the `tag`s that I want to output; this doesn't get into metadata, so I need to write it out some other way; how do I connect these up to the individuals in the tree sequence?

My proposed change just seems like it's providing access to something that's there already; if the problem is the name, we could define some other name for it (like, "treeSeqID")?

petrelharp:
From the email thread that prompted this:

> So assuming that works, that gets you identifiers on the .trees side.  On the VCF side, it's a good idea for each genome in the VCF file to be annotated with the same identifiers, but that presently is not done.  Where would such per-sample information typically be put, in a VCF file?  I'm looking at the VCF 4 spec and not seeing an obvious spot for per-sample annotations, which perhaps is why I didn't already do this (but probably I'm just missing the obvious).  :->  If you can suggest a good way for it to be encoded in the VCF, I could check that improvement in on GitHub within a day or two.

The last line in the header has the sample IDs, which you currently fill out like `i0 i1 i2 ...`. These could be replaced with like `i<pedigree ID>`...  if these are defined. This would be a good idea, I think.

bhaller:
That would still provide only the *individual* IDs, whereas it is likely the genome IDs that people want to get at.  I think we probably just want to output both the genome IDs and the individual IDs to their own lines in the VCF file; I just need to figure out the right way to do that.  Sorry that I'm being somewhat slow responding to these threads, I'm kinda sorta on vacation right now.  :->

I don't want to change the name for backward compatibility reasons; and I don't want to make the pedigree IDs user-visible when pedigree tracking has not been enabled.  Enabling pedigree tracking is a simple matter of flipping a switch, so that seems fine to me as a solution.  But feel free to argue your case more, I'm not issuing a final decision :->; I'll review this again when I'm back at work.

petrelharp:
>  feel free to argue your case more

OK! Well, we definately need a way to access the identifiers that get written out to the tree sequence from within SLiM, at the Eidos level. If you've a model where e.g., `tag`, or some `setValue`, is important, then we have to be able to write out that data in a way that it can be matched to the individuals in the tree sequence. Making pedigreeIDs available would do this, which seems very easy and natural, since the pedigreeIDs are showing up in the tree sequence, they ought to be visible in SLiM as well. Turning on pedigree recording whenever tree sequence recording is turned on is fine by me in principle, except that it adds overhead. If we can do some tests to show that this overhead is trivial then OK I guess, but it seems like it potentially opens up a bunch of unnecessary cans of worms. (For instance, when we reload from a tree sequence, the parental IDs won't be there any more.)

But, I'm probably missing something: what exactly is your objection to making pedigreeIDs available when pedigree recording is not turned on?

The question of VCFs is separate. I moved that to #43, sorry to clutter things up here.

> Sorry that I'm being somewhat slow responding to these threads

No worries! I know of no-one actually needing this yet, even: just thinking ahead.

bhaller:
This has been fixed as part of fixing #43, with slightly modified diffs, so I'm going to close this pull request.  Thanks, though; after reflection I came to agree with your position.  Turning on tree-seq recording does not "officially" enable pedigree tracking, and pedigree tracking features like parent/grandparent IDs and relatedness() calculations are not available, but it does now "officially" enable the individual and genome pedigree IDs themselves, making them accessible.

petrelharp:
I reviewed the changes; they look good.

PR #52, opened 2019-08-02
Qtslim

brnorris03:
CMakeLists.txt -- added include path to QtSLiM headers, attempting to fix error in failed build [#695.2](https://travis-ci.org/MesserLab/SLiM/builds/566742162).


bhaller:
Looks like it succeeded on Linux, so that's great.  It failed on macOS; it looks like maybe the way the .travis.yaml file installs Qt5 didn't set things up right, or something.  If you want to delve into that, that'd be great – it'd be good to have it working in the long term – but if you want to move on with your life, I can just disable the macOS build on Travis-CI for QtSLiM for now, it's not a big deal.  Up to you; do you want to pursue this now, or postpone it?  Either way, thanks for your help.  :->

brnorris03:
I can look into this, but not sure how long it will take -- don't have much experience with Qt5 on Mac, but have built other nasty things. I see you used brew for qt5, so I can try to replicate this on one of my macs and suggest a fix.

bhaller:
I used brew for no good reason; it's just a chunk of .travis.yaml script that I cribbed from some other project that also uses Qt5, so I was hoping it would work for me, but apparently not.  On my own machine I generally use Macports to install packages (for no particular reason – I've never seen a strong reason to prefer Macports over Homebrew or vice versa), but I believe, IIRC, I installed Qt5 with a double-click installer that I downloaded from their website, not using Macports.

bhaller:
Anyway, I'll wait a while to see whether you make progress before I take this PR, I guess.  It's good to see the Linux build succeed once on Travis; that's enough to make me happy for now.  :->

brnorris03:
I also use macports (I do have a good reason -- it was there before brew :), just never saw a reason to switch. My own Qt5 installs are also the OSX specific ones, but I can temporarily add a brew one. While it's not recommended to have macports and brew on the same machine, I have done this for testing without any issues. They don't share paths, so I don't really see the harm. I am swamped with paper reviews, but if I finish reasonably early, I will play with the Qt5 install (then offline through Wednesday!).

PR #53, opened 2019-08-05
Changed structure of building sparse arrays

suds12:
A few lines were meant only for my local copy but accidentally got pushed(like intel annotations). I will clean those out.


core/sparse_array.cpp
```diff
@@ -46,6 +47,25 @@ SparseArray::SparseArray(unsigned int p_nrows, unsigned int p_ncols)
...
+    nnz_capacity = (uint32_t *)malloc((nrows_) * sizeof(uint32_t));
+
+    for(int i=0;i<nrows_;i++)
+    {   
+        columns[i]=(uint32_t *)malloc((initial_width) * sizeof(uint32_t));
+        distances[i]=(uint32_t *)malloc((initial_width) * sizeof(uint32_t));
+    
+    //initial nnz and its capacity
+        nnz[i]=0;
+        nnz_capacity[i]=0;
```
petrelharp: this should be `initial_width`

core/sparse_array.cpp
```diff
@@ -46,6 +47,25 @@ SparseArray::SparseArray(unsigned int p_nrows, unsigned int p_ncols)
 	
 	row_offsets_[nrows_set_] = 0;
 	finished_ = false;
+
+    //Allocate space for 2-D vectors
+    //Columns and Distances are now 2-D vectors with each row corresponding to one focal individual. This negates the need for row offsets.
+
+    initial_width=5000;       //Some random initial width for each row. Can be changed later.
```
petrelharp: I'd say start with 100, then double it when you run out.

bhaller: Yes; the width of a row is the number of other individuals that one focal individual interacts with, and that will often be low even for large model, if the interaction distance is short.  Right now all of these buffers are re-allocated in every generation and grow up to their target size from scratch, over and over as the model runs.  Ideally the grown buffers, at the size that turns out to be sufficient for the model's dynamics, would be kept and re-used over time, so that once the code equilibrates no buffers ever need to be allocated or grown again.  Not sure whether that additional complication would be worth the work, though; we'll have to see whether malloc/realloc show up as significant overhead with this redesign.  For now 100 seems reasonable – but for some models it might already be too high, and might substantially increase the overhead for the spare array (which can be the bulk of all memory usage for large spatial models, I think).

core/sparse_array.cpp
```diff
@@ -68,6 +89,31 @@ SparseArray::~SparseArray(void)
 	
 	free(strengths_);
 	strengths_ = nullptr;
+
+    //Deallocating all multithreading parts 
+    for(int i=0;i<nrows_;i++)
+    {
+
+        free(columns[i]);
+        columns = nullptr;
```
bhaller: should be columns[i] = nullptr, no?

core/sparse_array.cpp
```diff
@@ -68,6 +89,31 @@ SparseArray::~SparseArray(void)
...
+
+        free(distances[i]);
+        distances = nullptr;
```
bhaller: likewise, distances[i] = nullptr

core/sparse_array.cpp
```diff
@@ -95,6 +141,31 @@ void SparseArray::Reset(unsigned int p_nrows, unsigned int p_ncols)
 	finished_ = false;
 }
 
+void SparseArray::Resize(uint32_t p_row)
+{
+    //vectors are reallocated by 1000 entries at a time. It can be wasteful but probably better than making a function call for each new entry. Will defer the final decision to Ben.
```
bhaller: Doubling with each realloc as Peter suggested is usually a good strategy.  I think that's what most C++ standard library implementations do too.  If you start with an initial size of zero, as I suggest above, then the initial width would be enforced here, with something like `new_width = (current_width == 0) ? initial_width : current_width*2;`.

core/sparse_array.cpp
```diff
@@ -95,6 +141,31 @@ void SparseArray::Reset(unsigned int p_nrows, unsigned int p_ncols)
 	finished_ = false;
 }
 
+void SparseArray::Resize(uint32_t p_row)
+{
+    //vectors are reallocated by 1000 entries at a time. It can be wasteful but probably better than making a function call for each new entry. Will defer the final decision to Ben. 
+
+
+    *(columns + p_row)=(uint32_t *)realloc(*(columns + p_row), 1000 * sizeof(uint32_t));
```
bhaller: This should be adding to the existing size, not just reallocing at a fixed size of 1000*sizeof(uint32_t), right?  Probably the only reason this code doesn't crash is that the model you're testing with probably never causes a realloc.  :->

core/sparse_array.cpp
```diff
@@ -95,6 +141,31 @@ void SparseArray::Reset(unsigned int p_nrows, unsigned int p_ncols)
 	finished_ = false;
 }
 
+void SparseArray::Resize(uint32_t p_row)
+{
+    //vectors are reallocated by 1000 entries at a time. It can be wasteful but probably better than making a function call for each new entry. Will defer the final decision to Ben. 
+
+
+    *(columns + p_row)=(uint32_t *)realloc(*(columns + p_row), 1000 * sizeof(uint32_t));  
+    *(distances + p_row)=(uint32_t *)realloc(*(distances + p_row), 1000 * sizeof(uint32_t));
```
bhaller: ditto

core/sparse_array.cpp
bhaller: You know, rather than pre-allocating all the rows at some initial width, you could actually just leave columns[i], distances[i], etc. as nullptr, it seems to me, and have nnz_capacity[i] be zero as you have now.  That would have the effect of deferring allocation of the row until the point when the row is written to for the first time, and would allow that allocation to be done by the threads rather than in a single-threaded block up front.  It would also have the benefit that if a row remains empty (because a given individual does not interact with any other individual) the row's buffers never get allocated at all – a savings in time and memory for models where that is common.  I think that design will probably be both simpler (all allocation is done by realloc) and potentially more efficient.

core/sparse_array.cpp
```diff
@@ -95,6 +141,31 @@ void SparseArray::Reset(unsigned int p_nrows, unsigned int p_ncols)
...
+
+    *(nnz_capacity + p_row)+= 1000; //Update capacity 
+}
+
+void SparseArray::AddDistance(uint32_t p_row, const uint32_t p_column, sa_distance_t p_distance)
```
bhaller: I'm not sure whether the compiler is already set up to inline methods that aren't marked explicitly as inline, but this is obviously a prime candidate for inlining; might move it to the header and make it explicitly inline to encourage dumb compilers to do the right thing.

core/sparse_array.cpp
suds12: Done!

core/sparse_array.cpp
```diff
@@ -49,18 +49,18 @@ SparseArray::SparseArray(unsigned int p_nrows, unsigned int p_ncols)
 	finished_ = false;
 
     //Allocate space for 2-D vectors
-    //Columns and Distances are now 2-D vectors with each row corresponding to one focal individual. This negates the need for row offsets.
+    //Columns and Distances are now ragged 2-D vectors with each row corresponding to one focal individual. This negates the need for row offsets.
 
-    initial_width=5000;       //Some random initial width for each row. Can be changed later.
+    initial_width=100;       //Some random initial width for each row. Can be changed later.
```
suds12: I realize 5000 is overkill.  Especially for mosquitos.

core/sparse_array.cpp
```diff
@@ -145,26 +145,13 @@ void SparseArray::Resize(uint32_t p_row)
 {
     //vectors are reallocated by 1000 entries at a time. It can be wasteful but probably better than making a function call for each new entry. Will defer the final decision to Ben. 
 
+    *(nnz_capacity + p_row)= (*(nnz_capacity + p_row)==0) ? initial_width : *(nnz_capacity + p_row) * 2; //Initialize or double the capacity 
 
-    *(columns + p_row)=(uint32_t *)realloc(*(columns + p_row), 1000 * sizeof(uint32_t));  
-    *(distances + p_row)=(uint32_t *)realloc(*(distances + p_row), 1000 * sizeof(uint32_t));
+    *(columns + p_row)=(uint32_t *)realloc(*(columns + p_row), *(nnz_capacity + p_row) * sizeof(uint32_t));  
+    *(distances + p_row)=(sa_distance_t *)realloc(*(distances + p_row), *(nnz_capacity + p_row) * sizeof(sa_distance_t));
```
suds12: I missed the fact that distances had their own struct and initialized with int. Took care of that.

core/sparse_array.cpp
suds12: Thanks for that. In general, that looks like the better way of creating 2d vectors. I've always created arrays first to some initial usage and then worry about reallocating. A relatively bad habit that I would rectify.

CMakeLists.txt
```diff
@@ -50,17 +55,27 @@ if(NOT CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 
 # Use the flags below for [all / Debug / Release] builds; these flags are built in to cmake
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-deprecated-register -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wl,--no-as-needed -Bdynamic -ldl -I /packages/intel/advisor/advisor_2019.4.0.597843/include/ -std=c11 -Wno-deprecated-register -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--no-as-needed -ldl -I /packages/intel/advisor/advisor_2019.4.0.597843/include/ -Bdynamic -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value")
```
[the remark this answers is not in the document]
suds12: -wl, --no-as-needed, -ldl, -Bdynamic are flags needed for source code annotations vtunes amplifier(a profiler). I will definitely clean them out before we merge to master but don't mind them hanging around in the openmp branch until then if thats ok with you. They don't mess with the actual program

petrelharp:
This is out of date - @suds12 has made more changes on this branch, and is going to start a new PR to discuss them. We can close this.

PR #69, opened 2020-01-25
Parallel SLiM 

suds12:


core/sparse_array.cpp
```diff
@@ -46,6 +46,25 @@ SparseArray::SparseArray(unsigned int p_nrows, unsigned int p_ncols)
 	
 	row_offsets_[nrows_set_] = 0;
 	finished_ = false;
+
+    //Allocate space for 2-D vectors
+    //Columns and Distances are now ragged 2-D vectors with each row corresponding to one focal individual. This negates the need for row offsets.
+
+    initial_width=100;       //Some random initial width for each row. Can be changed later.
```
petrelharp: We may want to start with a smaller number?

core/sparse_array.cpp
```diff
@@ -68,6 +88,31 @@ SparseArray::~SparseArray(void)
 	
 	free(strengths_);
 	strengths_ = nullptr;
+
```
petrelharp: Proposal: re-use this stuff: don't free/alloc, instead just clear the memory and reuse the data structure.

petrelharp:
To-do:

- [ ] take the changes to table sorting (from tskit) out, and put in a different PR
- [ ] write a bullet-pointed general overview of the changes
- [ ] improve description of how to build in README.md: for instance, isn't the directory `icc_build` only if using `icc`? what file are we supposed to change lines in? explain what's going on?
- [x] squash commits
- [ ] make Travis checks pass
- [ ] normalizing indentation to match Ben's
- [ ] Check how much the new sparse array data structure increases memory usage (hopefully not much: worst-case depends on the initial size and increase).
- [ ] Check how the size of the sparse array data structure grows with time (hopefully it doesn't grow after a few generations).
- [ ] rebase to master
- [ ] draft something for the changelog in VERSIONS

CMakeLists.txt
```diff
@@ -93,10 +110,12 @@ target_include_directories(${TARGET_NAME} PUBLIC)
 # TSKIT
 set(TARGET_NAME tables)
 file(GLOB_RECURSE TABLE_SOURCES ${PROJECT_SOURCE_DIR}/treerec/tskit/*.c)
+file(GLOB_RECURSE PARALLEL_TABLE_SOURCES ${PROJECT_SOURCE_DIR}/treerec/tskit/*.cpp)
 set(TSKIT_INCLUDES ${PROJECT_SOURCE_DIR}/treerec)
-add_library(${TARGET_NAME} STATIC ${TABLE_SOURCES})
+add_library(${TARGET_NAME} STATIC ${TABLE_SOURCES} ${PARALLEL_TABLE_SOURCES})
```
bhaller: Again, should only happen when the build is parallel; the cmake changes need to be completely isolated so that a standard build is unchanged

CMakeLists.txt
```diff
@@ -113,4 +132,4 @@ target_include_directories(${TARGET_NAME} PRIVATE ${GSL_INCLUDES} "${PROJECT_SOU
 target_link_libraries(${TARGET_NAME} PUBLIC gsl)
 target_link_libraries(${TARGET_NAME} PUBLIC tables)
 
-install(TARGETS slim eidos DESTINATION bin)
+install(TARGETS slim eidos DESTINATION /home/users/ssriniv2/local/parallel_slim)
```
bhaller: obviously this is not the install destination that we want in GitHub :->

EidosScribe/EidosHelpController.mm
```diff
@@ -158,6 +158,10 @@ - (void)dealloc
 // now substitutes an EidosNSString for the original NSString.  That EidosNSString responds to copyWithZone: by making a copy
 // of itself, preventing NSTaggedPointerString from getting in.  This seems to work, and I think it ought to be fairly safe
 // and airtight.  We shall see.  BCH 2/26/2018
+// BCH 11/27/2019: Note that the Qt version of this class now has a much nicer design, in which a proper tree class is used
```
bhaller: These are changes that occurred outside of your work, and are already committed to the openmp branch, so there seems to be a problem with the rebasing of this PR, or something.  I have just released SLiM 3.3.2, and just pulled over those changes into the openmp branch, so you ought to rebase onto what is not the head of the openmp branch for cleanliness.  Sorry things keep shifting; life working on a branch.

README.md
```diff
@@ -64,3 +64,19 @@ To build the code with debug flags turned on (i.e., optimization turned off, deb
 This could be done in a separate directory to retain both versions of the build.
 
 See the SLiM manual for more information about building and installing, including instructions on building SLiMgui under Xcode.  The manual and other SLiM stuff can be found at [http://messerlab.org/slim/](http://messerlab.org/slim/).
+
+Instructions for parallel implementation
+----------------------------------
+-Switch to openmp branch
+```
+cd icc_build
```
bhaller: Why are these build instructions different from the build instructions given in CMakeLists.txt?

README.md
```diff
@@ -64,3 +64,19 @@ To build the code with debug flags turned on (i.e., optimization turned off, deb
...
+source build
+make -j
+make install
+```
+- To toggle between serial and parallel builds: <br/>
+	-Remove line 21 : -DParallel=ON ..\\ in icc_build/build <br/>
+	-Change line line 20 to : -D CMAKE_BUILD_TYPE=Debug ..\\ <br/>
+	-Delete CMakeCache.txt <br/>
+	-Rebuild project<br/>
+- This is currently implemented in gcc but also works with icc(intel). To switch compilers, do so in icc_build/build
```
bhaller: This is all quite confusing.  If it is "currently implemented in gcc", why do the build instructions tell you to go to a directory called "icc_build"?  And why are there instructions here for modifying the cmake file, given that the parallel flag can be supplied to cmake on the command line?

SLiMgui/SLiMWindowController.mm
```diff
@@ -3695,7 +3695,7 @@ - (void)haplotypeProgressTaskFinished
 		
 		// then unique by pointer value to get a list of unique signatures (which may not be unique by name)
 		auto unique_end_iter = std::unique(propertySignatures->begin(), propertySignatures->end());
-		propertySignatures->resize(std::distance(propertySignatures->begin(), unique_end_iter));
+		propertySignatures->resize(static_cast<size_t>(std::distance(propertySignatures->begin(), unique_end_iter)));
```
bhaller: The next several changes again appear to be problems with the rebase.

core/interaction_type.cpp
```diff
@@ -24,11 +24,14 @@
...
 #include "slim_sim.h"
-
+#include <omp.h>
 #include <utility>
 #include <algorithm>
 
+#include <ctime>
+#include <chrono>
 
+using namespace std::chrono;
```
bhaller: I try to avoid "using" entirely, it is just asking for weird problems.  Just put "std::chrono::" where it's needed.

core/interaction_type.cpp
```diff
@@ -519,22 +522,38 @@ void InteractionType::CalculateAllDistances(Subpopulation *p_subpop)
...
 					case 1:
 						for (row = start_row; row < after_end_row; row++)
+                        {
 							BuildSA_1(subpop_data.kd_root_, position_data + row * SLIM_MAX_DIMENSIONALITY, row, subpop_data.dist_str_);
+                        }
 						break;
 					case 2:
-						for (row = start_row; row < after_end_row; row++)
+                    //ANNOTATE_SITE_BEGIN(s7)
+#ifdef Parallel
```
bhaller: I thought OMP's design was such that these lines did not need to be #ifdef'd, because when not doing a parallel build the omp.h header had stubs in place that rendered all of the parallel constructs no-ops.  No?

core/main.cpp
```diff
@@ -156,6 +156,8 @@ static void test_exit(int test_result)
 
 int main(int argc, char *argv[])
 {
+    //std::cout<<"started main \n";
+    // double start_time = omp_get_wtime();                   //Start timer
```
bhaller: Note that the -t command-line flag to SLiM already provides a summary of total runtime.  But maybe omp_get_wtime() is providing you with different information than that provides?

core/mutation.h
```diff
@@ -69,7 +69,8 @@ class Mutation : public SLiMEidosDictionary
 	slim_objectid_t subpop_index_;						// subpopulation in which mutation arose (or a user-defined tag value!)
 	const slim_generation_t origin_generation_;			// generation in which mutation arose
 	int8_t nucleotide_;									// the nucleotide being kept: A=0, C=1, G=2, T=3.  -1 is used to indicate non-nucleotide-based.
-	// NOTE THERE ARE 3 BYTES FREE IN THE CLASS LAYOUT HERE; see Mutation::Mutation() and Mutation layout.graffle
+	int8_t scratch_;									// temporary scratch space for use by algorithms; regard as volatile outside your own code block
```
bhaller: More rebasing problems

core/slim_sim.cpp
```diff
@@ -4722,51 +4723,52 @@ void SLiMSim::ReorderIndividualTable(tsk_table_collection_t *p_tables, std::vect
...
 
+
 void SLiMSim::SimplifyTreeSequence(void)
 {
 #if DEBUG
 	if (!recording_tree_)
 		EIDOS_TERMINATION << "ERROR (SLiMSim::SimplifyTreeSequence): (internal error) tree sequence recording method called with recording off." << EidosTerminate();
 #endif
-	
+
```
bhaller: please try to get rid of "noise" like this in the PR, it makes it hard to review

core/sparse_array.cpp
bhaller: Presumably the number should be chosen with some kind of empirical basis, although the optimal number probably varies depending on the platform and the model; still, we should at least have some empirical sense of whether 10 or 1000 is better...

core/sparse_array.cpp
```diff
@@ -46,6 +46,25 @@ SparseArray::SparseArray(unsigned int p_nrows, unsigned int p_ncols)
 	
 	row_offsets_[nrows_set_] = 0;
 	finished_ = false;
+
+    //Allocate space for 2-D vectors
+    //Columns and Distances are now ragged 2-D vectors with each row corresponding to one focal individual. This negates the need for row offsets.
```
bhaller: Is this a change we want to make only in the parallel case, or a change that will be made for the single-threaded case also?  What are the performance implications of it for the single-threaded case?  Presumably negative; but how negative?

core/sparse_array.cpp
```diff
@@ -95,6 +140,18 @@ void SparseArray::Reset(unsigned int p_nrows, unsigned int p_ncols)
 	finished_ = false;
 }
 
+void SparseArray::Resize(uint32_t p_row)
+{
+    //vectors are reallocated by 1000 entries at a time. It can be wasteful but probably better than making a function call for each new entry. Will defer the final decision to Ben.
```
bhaller: This comment says "by 1000 entries at a time", but the code doubles the capacity with each realloc.  Doubling is probably reasonable; you know you're not wasting more than 50% of memory in the worst case, and it avoids a large number of reallocs if the final size is >> initial_width.  Most STL implementations of things like std::vector double the capacity when it is exceeded, as I understand it.

core/sparse_array.h
```diff
@@ -57,12 +57,21 @@ class SparseArray
 	
 	uint32_t nrows_, ncols_;		// the number of rows and columns; determined at construction time
 	uint32_t nrows_set_;			// the number of rows that have been configured (at least partially, during building)
-	uint32_t nnz_;					// the number of non-zero entries in the sparse array (also at row_offsets[nrows_set])
 	uint32_t nnz_capacity_;			// the number of non-zero entries allocated for at present
 	
 	bool finished_;					// if true, Finished() has been called and the sparse array is ready to use
+
+
+	uint32_t nnz_;					// the number of non-zero entries in the sparse array (also at row_offsets[nrows_set])
```
bhaller: Seems like this line was better where it used to be, above...?

core/sparse_array.h
```diff
@@ -123,7 +132,26 @@ class SparseArray
 		columns_[offset] = p_column;
 		distances_[offset] = p_distance;
 	}
-	void AddEntryInteraction(uint32_t p_row, const uint32_t p_column, sa_distance_t p_distance, sa_strength_t p_strength);
+   
+    //Added by Sudharshan    
+    void Resize(uint32_t p_row);
```
bhaller: Maybe "ExpandRow" or "IncreaseRowCapacity" or something would be more clear?

eidos/eidos_globals.cpp
```diff
@@ -1880,25 +1879,86 @@ double Eidos_ExactSum(const double *p_double_vec, int64_t p_vec_length)
 	return hi;
 }
 
-// Thanks to user Sviatoslav at https://stackoverflow.com/a/37454181/2752221 for this code
-std::vector<std::string> Eidos_string_split(const std::string &p_str, const std::string &p_delim)
```
bhaller: Lots more rebase noise here...

treerec/tskit/parallel_sort.cpp
```diff
@@ -0,0 +1,106 @@
...
+//#include "tbb/parallel_sort.h"
+
+extern "C"{
+#endif
+
+void psort_edges(edge_sort_t *sorted_edges, tsk_size_t n);
+void psort_sites(tsk_site_t *sorted_sites, tsk_size_t num_sites);
+void psort_mutations(tsk_mutation_t *sorted_mutations, tsk_size_t num_mutations);
+
+//Bool based comparator functions for sorting. The old ones were giving me problems so recoded it. Hope that's fine.
```
bhaller: Probably fine; ideally this sort of thing will get pushed back into tskit, though.  Things that are C++ frosting on top of tskit maybe should be put in a new and different place, outside of the tskit/ subdirectory; @petrelharp?

treerec/tskit/parallel_sort.cpp
```diff
@@ -0,0 +1,106 @@
...
+        /* Within a particular site sort by ID. This ensures that relative ordering
+         * within a site is maintained */
+        return a.id < b.id;
+    }
+}
+
+void psort_edges(edge_sort_t *sorted_edges, tsk_size_t n)
+{
+#ifdef Parallel
+    __gnu_parallel::sort(sorted_edges, sorted_edges+n ,&cmp_edges);
```
bhaller: So, my recollection was that the use of GNU's parallel sort was just a temporary thing for exploration, but that we were going to use some non-GNU solution (perhaps a parallel sort function that we just write ourselves, or borrow from GNU or Boost or whatever) to avoid the external dependency...?  This should probably be resolved before this PR goes in...?  I'm not sure that we need the openmp branch to be in a final state, ready to ship; that's probably setting the bar too high.  But I'm not sure we want to be committing experimental things that need to be changed, either; that feels like maybe setting the bar too low.  Thoughts?

treerec/tskit/parallel_sort.hpp
```diff
@@ -0,0 +1,13 @@
+#ifdef __cplusplus
+#include <tskit/tables.h>
+extern "C"{
+#endif 
+
+
+template<typename T>
```
bhaller: Is this template live code?  It looks like you've now got separate implementations for psort_mutations(), psort_sites(), etc.; but I'm not sure I'm fully understanding the diffs.  If this is obsolete code, please clear it out.

treerec/tskit/tables.c
```diff
@@ -3399,14 +3399,7 @@ tsk_provenance_table_load(tsk_provenance_table_t *self, kastore_t *store)
...
  *************************/
-
-typedef struct {
-    double left;
-    double right;
-    tsk_id_t parent;
-    tsk_id_t child;
-    double time;
-} edge_sort_t;
+//Moved Struct edge_sort_t to tables.h(Sudharshan)
```
bhaller: Again I wonder: what's the long-term strategy here, @petrelharp?  Upon reflection, it seems like it would be a Good Thing for tskit to have some awareness of parallelism; we're presumably not going to be the only tskit client that will want to do these things, and so maybe we ought to plan for the parallelism changes in tskit in SLiM to ultimately get pulled back into tskit itself.  But I guess some of it is C++-based now (does it need to be, or could it be pure C?); is that a total obstacle to it being in tskit, or do you think Jerome might accept a C++ veneer in the tskit sources that would be an optional add-on in some way?  Maybe we need to discuss our approach on this a bit.

treerec/tskit/tables.c
```diff
@@ -3419,57 +3412,57 @@ typedef struct {
...
-            ret = (ca->child > cb->child) - (ca->child < cb->child);
-            /* If the child nodes are equal, sort by the left coordinate. */
-            if (ret == 0) {
-                ret = (ca->left > cb->left) - (ca->left < cb->left);
-            }
-        }
-    }
-    return ret;
-}
+//The following 3 comparator functions are replaced by new bool based  ones in parallel_sort.cpp
```
bhaller: So, bool-based comparators are not actually quite the same, right?  The int comparators return -1, 0, or 1, where 0 means "compared equal"; the bool-based comparators don't have that option.  What are the implications?  Will the sort order be subtly different when using the new comparators?  That could introduce some very subtle bugs that would have us tearing our hair out.  Is this change absolutely necessary?  @petrelharp, thoughts?

treerec/tskit/tables.c
```diff
@@ -3523,7 +3516,10 @@ table_sorter_sort_edges(table_sorter_t *self, tsk_size_t start)
         e->child = self->edges->child[k];
         e->time = self->nodes->time[e->parent];
     }
-    qsort(sorted_edges, n, sizeof(edge_sort_t), cmp_edge);
+    
+    //Commented out quick sort(serial) for now. Will add switch for serial execution once Ben verifies this part
```
bhaller: Might as well add this switch; I'm not sure all the details are sorted out yet, but let's try to get the code into as final a state as we can.  No harm in adding the #ifdef here and calling qsort when we're single-threaded (if that's what you plan to do); it will make the overall design plan clearer.

treerec/tskit/tables.h
```diff
@@ -103,6 +102,16 @@ typedef struct {
...
 } tsk_individual_t;
 
+//Moved from tables.c(sudharshan)
+typedef struct {
+    double left;
+    double right;
+    tsk_id_t parent;
+    tsk_id_t child;
+    double time;
+} edge_sort_t;
```
bhaller: I'm kinda scratching my head as to what this struct even is.  Why does the comparator for sorting the edge table use a special struct instead of just using the struct that the edge table actually uses?  I know this is tskit's design, not yours.  Just seems a bit odd.

bhaller: @petrelharp, do you understand why tskit does this?

treerec/tskit/tables.c
bhaller: I guess since qsort() and std::sort() are not guaranteed to be stable, perhaps this is a non-issue; our replacement sorting does not have to guarantee stability either...?  I do note that since the single-threaded case will presumably continue to use the old sorting strategy, so we presumably don't want to comment out the int-based comparators, but instead just add new bool-based comparators as well...?

CMakeLists.txt
```diff
@@ -93,10 +110,12 @@ target_include_directories(${TARGET_NAME} PUBLIC)
 # TSKIT
 set(TARGET_NAME tables)
 file(GLOB_RECURSE TABLE_SOURCES ${PROJECT_SOURCE_DIR}/treerec/tskit/*.c)
+file(GLOB_RECURSE PARALLEL_TABLE_SOURCES ${PROJECT_SOURCE_DIR}/treerec/tskit/*.cpp)
```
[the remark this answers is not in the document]
bhaller: See further comment on this issue below.

CMakeLists.txt
```diff
@@ -40,6 +40,11 @@
 # Then you can just do "make"; cmake will automatically be re-run by make since the CMakeLists.txt
 # file has changed.
 
+# By default, this will make a serial execution build but can be changed by turning on the parallel flag
+#
+#   mkdir Debug
+#   cd Debug
+#   Cmake -D Parallel=ON
```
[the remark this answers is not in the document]
bhaller: And since, IIRC, the default build is a Release build, if you're going to use a folder called "Debug" here then you ought to supply the `-D CMAKE_BUILD_TYPE=Debug` flag.  But probably you ought to show how to make a parallel Release build, not a Debug build, anyway; presumably that will be what most people want.

CMakeLists.txt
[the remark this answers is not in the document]
petrelharp: not sure - I need the big picture here?

treerec/tskit/parallel_sort.cpp
petrelharp: yes, we need to figure out the right way to do this

suds12:
https://docs.google.com/document/d/1jzjob289Yx36EKts49n3Uig-jywMFcHjmcLLioKQHwE/edit?usp=sharing

suds12:
- Overview of changes
- Set up instrumentation for building SLiM uzaing OpenMP.
- Changed structure of the spazrse array for maintaining interactions from CSR representation to 2-dimensional vectors that have dedicated columns for each row.
- With this change, each row can be built independently without having to wait for the previous row to be built unlike CSR, where each row is added after the previous row, thereby, needing the exact location to start building next row(which is not known until each row is finished).
- Reverted to recursive implementation of traversing KD tree(for building a sparse array) from its equivalent non-recursive implementation. The latter explicitly maintains a call stack for mimicking recursion. This stack is simultaneously being accessed by multiple threads during parallel execution and hence causes read after write dependency. This is avoided in a recursive implementation as the call stack is maintained internally by the compiler making it parallel safe.
- Parallelized building sparse array using OpenMP.
- Replaced qsort() in tree sequence simplification with psort(). This is a C++ wrapper function created for using parallel STL which is not provided for C.
- Replaced int based comparator function for sorting with a bool based one. This shouldn’t make a difference as the original qsort() is unstable either way.
- Parallelized sorting of edges, sites and mutations tables 
- Moved a few structures around in tskit to accommodate the new wrapper function.
- 
- -

petrelharp:
Add to the list:

- [ ] make a PR to [tskit](https://github.com/tskit-dev/tskit) that allows passing in of the sorting function to [tsk_table_collection_sort](https://github.com/tskit-dev/tskit/blob/a82bab156996b0728c0021a430ee8001712573c1/c/tskit/tables.c#L6525) - proposal is change the API to take a struct of options instead of just a (integer) flag.

petrelharp:
Thanks a ton for the comments, @bhaller - I haven't responded to your queries because we are discussing the right way to do the tskit integration, and it's going to be somewhat different (see above).  @suds12 has done a bunch of these things and will push these changes shortly.

treerec/tskit/tables.c
brnorris03: Regarding the comparison operator, @bhaller, what's your opinion about the standard <=> C++ operator? Is it appropriate to consider here?

CMakeLists.txt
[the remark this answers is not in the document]
suds12: It should work with both, but I changed it nonetheless

CMakeLists.txt
```diff
@@ -50,17 +55,29 @@ if(NOT CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 
 # Use the flags below for [all / Debug / Release] builds; these flags are built in to cmake
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-deprecated-register -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wl,--no-as-needed -Bdynamic -ldl -I /packages/intel/advisor/advisor_2019.4.0.597843/include/ -std=c11 -Wno-deprecated-register -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--no-as-needed -ldl -I /packages/intel/advisor/advisor_2019.4.0.597843/include/ -Bdynamic -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value")
+
 
-set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -g -Og -DDEBUG=1")
```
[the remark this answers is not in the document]
suds12: It is optimized debugging flag(o1 except for debugging symbols). https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html for reference

bhaller:
> Thanks a ton for the comments, @bhaller - I haven't responded to your queries because we are discussing the right way to do the tskit integration, and it's going to be somewhat different (see above). @suds12 has done a bunch of these things and will push these changes shortly.

OK.  Let me know when it's ready for me to do another pass.

treerec/tskit/tables.c
bhaller: > Regarding the comparison operator, @bhaller, what's your opinion about the standard <=> C++ operator? Is it appropriate to consider here?

I'm afraid you overestimate my C++ knowledge, @brnorris03; I didn't know any such thing as a <=> operator existed in C++ until I googled it just now.  :->  I can delve into that question if necessary, but at present I have no knowledge to contribute.

CMakeLists.txt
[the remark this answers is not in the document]
bhaller: Ah, that's great, never saw it before.  I should start using it.  Don't know why I couldn't find this info with Google, except that googling strings like "-Og" doesn't work very well.  :->

treerec/tskit/tables.h
petrelharp: Yes - `time` is not stored by an `edge` (as seen [here](https://github.com/tskit-dev/tskit/blob/a82bab156996b0728c0021a430ee8001712573c1/c/tskit/tables.h#L139)); but we want to sort edges by time (I forget if this is parent or child time).

CMakeLists.txt
```diff
@@ -50,17 +55,29 @@ if(NOT CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 
 # Use the flags below for [all / Debug / Release] builds; these flags are built in to cmake
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-deprecated-register -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wl,--no-as-needed -Bdynamic -ldl -I /packages/intel/advisor/advisor_2019.4.0.597843/include/ -std=c11 -Wno-deprecated-register -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value")
```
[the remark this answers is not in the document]
suds12: Done!

CMakeLists.txt
suds12: Done!

core/main.cpp
suds12: I have removed this but the reason why I use omp_get_wtime() is that it tracks time better for multithreaded segments.

core/sparse_array.h
suds12: Done!

treerec/tskit/parallel_sort.hpp
suds12: Done!

CMakeLists.txt
suds12: Done!

README.md
suds12: Since I initially wanted to have tbb stuff, I had a whole set of initializations that I wrote in a build file. Now that I removed tbb, the instructions are the same as its serial counterpart with the only  change being, -D Parallel=ON

README.md
suds12: Read the above comment.

petrelharp:
**Note:** we have determined that

1. this PR will be merged into `openmp`
2. this PR will make it so that `openmp` will be "release quality", so that we can tell early adopters to use this work - in particular, we should be sure that everything works properly
3. we will maintain openmp as "release quality" moving forwards but with the caveat that it is experimental - how stuff works exactly might change
4. we will accumulate enough parallization improvements in `openmp` that it constitutes a big, impressive improvement to most models when we merge it in to master

bhaller:
> **Note:** we have determined that
> 
> 1. this PR will be merged into `openmp`
> 2. this PR will make it so that `openmp` will be "release quality", so that we can tell early adopters to use this work - in particular, we should be sure that everything works properly
> 3. we will maintain openmp as "release quality" moving forwards but with the caveat that it is experimental - how stuff works exactly might change
> 4. we will accumulate enough parallization improvements in `openmp` that it constitutes a big, impressive improvement to most models when we merge it in to master

Agreed on all points except that, as we discussed at the end, "release quality" for the openmp branch will mean something a bit different than it does for the master branch, until we have done the additional validation work that we agreed needs to be done.  We will be able to say "as far as we know it doesn't break anything", but until we do that additional validation work I don't think we will really have the same level of confidence in its correctness that I strive for with official SLiM releases.  (And even then, of course, bugs do slip through the cracks.)

petrelharp:
> "as far as we know it doesn't break anything"

Agreed, with the addition that "and we have good reason to think it doesn't break anything".

suds12:
To-do:
- [x] take the changes to table sorting (from tskit) out, and put in a different PR
- [ ] write a bullet-pointed general overview of the changes
- [x] improve description of how to build in README.md: for instance, isn't the directory `icc_build` only if using `icc`? what file are we supposed to change lines in? *Mostly: explain what's going on?* 
- [x] squash commits
- [ ] make Travis checks pass
- [x] normalizing indentation to match Ben's
- [ ] Check how much the new sparse array data structure increases memory usage (hopefully not much: worst-case depends on the initial size and increase).
- [ ] Check how the size of the sparse array data structure grows with time (hopefully it doesn't grow after a few generations).
- [x] rebase to master
- [x] draft something for the changelog in VERSIONS
- [x] move the stuff in `icc_build/build` to CMake configuration somehow, and explain about it at the bottom of README in a section called Troubleshooting (say)
- [x] remove tbb code
- [ ] make a PR to tskit that allows passing in of the sorting function to tsk_table_collection_sort - proposal is change the API to take a struct of options instead of just a (integer) flag.
- [x] explain somewhere how to run with X threads (e.g. a single thread)
- [ ] make additional tests in SLiM that really exercise parallelization

core/sparse_array.cpp
bhaller: This question remains important and unresolved.  If this new ragged-row design for SparseArray is significantly slower than the old design in the non-parallel case – which would not surprise me at all – then I think it would be justified to actually have two separate implementations of this class, one used when serial, the other used when parallel.  It is not a large class, and the distinction would be completely hidden behind a shared external API, so the messiness of it would be minimal.  We need speed testing to determine this.  This decision can/should be pended until the redesign of SparseArray is complete, though, and it passes the necessary validation tests.  Then we should test the speed impact for the serial case, and if it is significant, make the code duplication.

core/sparse_array.cpp
```diff
@@ -38,14 +38,32 @@ SparseArray::SparseArray(unsigned int p_nrows, unsigned int p_ncols)
...
+    //Allocate space for 2-D vectors
+    //Columns and Distances are now ragged 2-D vectors with each row corresponding to one focal individual. This negates the need for row offsets.
+
+    initial_width=100;       //Some random initial width for each row. Can be changed later.
+    columns = (uint32_t **)malloc((nrows_) * sizeof(uint32_t *));
+    distances = (sa_distance_t **)malloc((nrows_) * sizeof(sa_distance_t *));
+    nnz = (uint32_t *)malloc((nrows_) * sizeof(uint32_t));
+    nnz_capacity = (uint32_t *)malloc((nrows_) * sizeof(uint32_t));
+
+    for(int i=0;i<nrows_;i++)
```
bhaller: Much better than looping to zero out these vectors would be to use calloc() instead of malloc() above.  That lets the kernel zero out the memory for us, which can be *way* faster than doing it ourselves.  Whenever you write a loop to zero out memory, think about whether you could use calloc() (for a new allocation) or bzero (for an existing allocation).

core/sparse_array.cpp
bhaller: Yes, might be a good idea, if the allocations/frees show up as significant in a profile.  This is the destructor for SparseArray, though, so it pretty much has to deallocate the memory it was using (perhaps it could return it to a shared pool, but that might be more work than it is worth since the blocks are not of a set fixed size).  So if we want to get into this optimization, probably the right level for it to happen at would be reuse of the SparseArray object itself.  That could probably be done inside InteractionType.  Food for thought, depending on whether this shows up as a bottleneck.

core/sparse_array.cpp
```diff
@@ -68,6 +87,31 @@ SparseArray::~SparseArray(void)
...
+
+    //Deallocating all multithreading parts 
+    for(int i=0;i<nrows_;i++)
+    {
+
+        free(columns[i]);
+        columns[i] = nullptr;
+
+        free(distances[i]);
+        distances[i] = nullptr;
```
bhaller: I do generally like to zero out instance variables in dealloc, so that if somebody downstream tries to use the variables they crash.  Here, though, I wouldn't bother zeroing out columns[i] and distances[i], because we're about to free and zero columns and distances themselves, so these values will all be inaccessible momentarily anyway.  Save yourself nrows_*2 writes to memory, for what that's worth.

core/sparse_array.cpp
bhaller: The comment here remains inaccurate, as discussed above; please fix it to reflect the behavior of the code.

core/sparse_array.cpp
```diff
@@ -105,7 +161,7 @@ void SparseArray::_ResizeToFitNNZ(void)
 		
 		columns_ = (uint32_t *)realloc(columns_, nnz_capacity_ * sizeof(uint32_t));
 		distances_ = (sa_distance_t *)realloc(distances_, nnz_capacity_ * sizeof(sa_distance_t));
-		strengths_ = (sa_strength_t *)realloc(strengths_, nnz_capacity_ * sizeof(sa_strength_t));
+        strengths_ = (sa_strength_t *)realloc(strengths_, nnz_capacity_ * sizeof(sa_strength_t));
```
bhaller: whitespace change, delete

core/sparse_array.h
```diff
@@ -57,12 +57,18 @@ class SparseArray
...
 	bool finished_;					// if true, Finished() has been called and the sparse array is ready to use
+	uint32_t nnz_;					// the number of non-zero entries in the sparse array (also at row_offsets[nrows_set])
 	
-	void _ResizeToFitNNZ(void);
+    //Stuff for multithreading
+    uint32_t initial_width;
+    uint32_t **columns;
+    sa_distance_t **distances;
+    uint32_t *nnz;
+    uint32_t *nnz_capacity;
```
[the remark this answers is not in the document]
suds12: So this is precisely the discrepancy I was talking about that might be giving the Travis test error.  The old sparse array still exists because it is being used build _SA1() and SA3(). Technically, the remaining stuff is indeed "converting other parts of the code, outside of SparseArray, to use the new SparseArray design". I went ahead with changing only SA2() because that was the one targetted by the benchmark scripts and I wanted to be as less intrusive as possible to the original codebase for my first PR but I should have probably thought this through.

CMakeLists.txt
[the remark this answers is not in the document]
petrelharp: Note: Boyana has suggestions for changing these flags.

core/interaction_type.cpp
```diff
@@ -50,9 +52,10 @@ std::ostream& operator<<(std::ostream& p_out, IFType p_if_type)
...
-	sim_(p_sim), interaction_type_id_(p_interaction_type_id), spatiality_string_(p_spatiality_string), reciprocal_(p_reciprocal), max_distance_(p_max_distance), max_distance_sq_(p_max_distance * p_max_distance), receiver_sex_(p_receiver_sex), exerter_sex_(p_exerter_sex), if_type_(IFType::kFixed), if_param1_(1.0), if_param2_(0.0),
+	sim_(p_sim),
 	self_symbol_(Eidos_GlobalStringIDForString(SLiMEidosScript::IDStringWithPrefix('i', p_interaction_type_id)),
-				 EidosValue_SP(new (gEidosValuePool->AllocateChunk()) EidosValue_Object_singleton(this, gSLiM_InteractionType_Class)))
+			 EidosValue_SP(new (gEidosValuePool->AllocateChunk()) EidosValue_Object_singleton(this, gSLiM_InteractionType_Class))),
+	spatiality_string_(p_spatiality_string), reciprocal_(p_reciprocal), max_distance_(p_max_distance), max_distance_sq_(p_max_distance * p_max_distance), receiver_sex_(p_receiver_sex), exerter_sex_(p_exerter_sex), if_type_(IFType::kFixed), if_param1_(1.0), if_param2_(0.0), interaction_type_id_(p_interaction_type_id)
```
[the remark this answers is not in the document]
suds12: Well technically I never really changed this part. This version seems to be exactly the same as the one when I started updating the file. You might have changed it sometime in the past and when I merged with v3.3 in my last commit to remove noise, this probably got counted as a double change. Correct me if I am wrong. Sorry for that.

core/sparse_array.cpp
petrelharp: nice - summary is that (a) the current implementation ends up having a fair amount of unused memory, because it resizes by doubling; (b) the new implementation does a lot of different doublings instead of just one big one; (c) but this means that the new implementation does order N resizings, which is bad; (d) but we can avoid this by starting out at the maximum length from the previous generation in each generation

bhaller: > nice - summary is that (a) the current implementation ends up having a fair amount of unused memory, because it resizes by doubling; (b) the new implementation does a lot of different doublings instead of just one big one; (c) but this means that the new implementation does order N resizings, which is bad;

Since each resizing is a doubling, this is O(log n) doublings, no?

(d) but we can avoid this by starting out at the maximum length from the previous generation in each generation

Or just by reusing the SparseArray objects, as we've previously discussed, rather than disposing of it and starting from scratch each generation, no?  If we clear the SparseArray but zero out the internal data structures, once the model has equilibrated there should be no reallocs at all.

petrelharp: > Since each resizing is a doubling, this is O(log n) doublings, no?

Your implementation does log(N) doublings; but in the new implementation, every one of the N rows is resized by doubling until it gets long enough.

petrelharp: > Or just by reusing the SparseArray objects, as we've previously discussed, rather than disposing of it and starting from scratch each generation, no?

yes! we agree!

bhaller: > > Since each resizing is a doubling, this is O(log n) doublings, no?
> 
> Your implementation does log(N) doublings; but in the new implementation, every one of the N rows is resized by doubling until it gets long enough.

Ah, true; so O(N log N), then, I guess.

petrelharp: Well, O(N log (number of neighbors)), whatever that is.

bhaller:
Closing because this is superseded by https://github.com/MesserLab/SLiM/pull/97.

PR #75, opened 2020-02-23
Added the Eidos function qnorm() and accompanying test functions

vsbuffalo:
Here is a draft pull request of the Eidos `qnorm()` function. Specifically, it

- Adds the necessary GSL code to use its function `gsl_cdf_gaussian_Pinv()`.
- Implements an Eidos version of qnorm(), that is vectorized like dnorm().
- Adds tests including exception testing.

The function works as expected as far as I can tell. I want to confirm you agree (1) the unit tests look sufficient (true values calculated against R's `qnorm()`) and (2) the exception handling for probabilities out of bounds look good. 

I'll begin working on documentation I can send your way in a bit.


bhaller:
Looks very clean to me.  Once Travis-CI gives it a thumbs-up and I've got the doc for it from you, I'll accept the pull.  Thanks!

vsbuffalo:
Thanks! Here's the documentation blurb. 

In the next few days, I'll code up a truncation selection example, which I can share as well if that's useful.

```
(float)qnorm(float p, [numeric mean = 0], [numeric sd = 1])
Returns a vector of quantiles for a normal distribution with lower tail probabilities less than p, with mean mean and standard deviation sd. The mean and sd parameters may either be singletons, specifying a single value to be used for all of the quantiles, or they may be vectors of the same length as p, specifying a value for each quantile computation.
```

bhaller:
Sounds good.  There's a truncation selection example in section 13.6 of the current manual; I'll be curious to see how yours differs and what you're using this new function for, exactly.  If you want to contribute an additional chunk to that section 13.6 stuff, that would be welcome (assuming it's sufficiently different from what's there already).

I'll integrate the new doc into the project in a little while.  Thanks a bunch!  It's nice to get external contributions like this.  :->

PR #104, opened 2020-07-31
skeleton of R tests for eidos functions

petrelharp:
Here's an experient with using R to make some unit tests for Eidos functions.  Execute `run_tests.sh` to run the tests - neither of which pass, because (a) of rounding/printing, and (b) quantile does not agree.

Note that at present, it doesn't *tell* you things don't pass - you have to look at the logs to see this. Work in progress.

Maybe this is useful, maybe it's not? It'd be real easy to extend this to test all the numerical Eidos methods, at least.

### DIFF
--- a/tests/R_tests/run_tests.sh
+++ b/tests/R_tests/run_tests.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+for test in test_*.R
+do
+    Rscript $test &>${test%.R}.log
+done
--- a/tests/R_tests/test_mean.R
+++ b/tests/R_tests/test_mean.R
@@ -1,0 +1,16 @@
+source("util.R")
+set.seed(23)
+
+method <- "mean"
+verbose <- FALSE
+
+for (x in numeric_vectors(10, min_length=1)) {
+    if (verbose) {
+        cat("::::::: doing: \n")
+        cat(strify(x))
+        cat("\n")
+    }
+    if(!compare_numeric(method, x)) {
+        stop(paste("Error!", method, "does not agree on", strify(x)))
+    }
+}
--- a/tests/R_tests/test_quantile.R
+++ b/tests/R_tests/test_quantile.R
@@ -1,0 +1,27 @@
+source("util.R")
+set.seed(23)
+
+method <- "quantile"
+verbose <- TRUE
+
+problist <- list(
+                 0.0,
+                 0.5,
+                 0.99999,
+                 c(0, 1/pi, 0.5, 0.9999, 1.0),
+                 seq(0, 1, length.out=5),
+                 seq(0, 1, length.out=100)
+                 )
+
+for (x in numeric_vectors(10, min_length=1)) {
+    for (probs in problist) {
+        if (verbose) {
+            cat(paste("::::::: doing:", method, "\n"))
+            cat(paste("x =", strify(x), "\n"))
+            cat(paste("probs =", strify(probs), "\n"))
+        }
+        if(!compare_numeric(method, x, probs)) {
+            stop(paste("Error!", method, "does not agree on", strify(x)))
+        }
+    }
+}
--- a/tests/R_tests/util.R
+++ b/tests/R_tests/util.R
@@ -1,0 +1,91 @@
+###
+# Things for constructing eidos tests
+
+# change this to the absolute path of a different binary to test with that one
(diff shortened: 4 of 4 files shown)


bhaller:
Hmm.  If this were to really be useful, it'd have to get quite a bit more complicated, I suspect.  Here are some issues I can think of off the top of my head:

- rounding, as you see; but that is not an easy thing to truly solve, because you *do* want to catch numerical/algorithmic problems that result in an unnecessary loss of precision in a computation, but you *don't* want to just flag all differences between R and Eidos in the n'th decimal place as being bugs...
- for Eidos, to exercise all the code paths, you often need to test several length cases: empty vector, singleton, non-singleton (these are very often handled by different code paths); for a function or operator that takes two parameters/operands, you need to test all combinations of those, like 1+1, 1+(1:5), (1:5)+1, (1:5)+(1:5), integer(0)+1, etc.
- depending on the function, you would want to test integer, float, logical, and string; in some cases NULL can be mixed in (or if it can't, then you want to check that you get an error); for float you also want to check results for computations involving NAN, INF, and -INF; some functions/operators will let you mix types together (integer and float, very commonly), and that would also need separate testing since it is often a different code path
- the results between R and Eidos are not identical in all cases; for example, operations in R that would produce NA have to either produce NAN or an error in Eidos since Eidos doesn't have NA (so I made decisions, on a case-by-case basis, about which behavior any given operation would have); and there are some other corner cases where I chose to make Eidos different from R (for example, Eidos is much less tolerant of things like vectors of mismatched length than R is, and Eidos strives to give integer results from integer operands where possible where R usually just throws you over to numeric type); so the test harness can't just assume "a difference between R and Eidos is a bug", it has to understand the differences between the languages
- in some cases the results from R and Eidos may mismatch in ways that are fine; for example, the results of setIntersection() in Eidos and intersect() in R might be in a different order, but that is (probably) fine since sets are unordered; but in other cases the order of the result vector does matter and must match; so the test harness needs to understand those semantics too
- then there are matrices and arrays; R and Eidos should handle them in fairly parallel ways, but all of the above sorts of caveats apply in this area as well, in all kinds of different ways (matrix mixed with singleton, singleton mixed with matrix, integer matrix mixed with float matrix, float matrix containing a NAN... the mind boggles).

So... I don't know.  There's a reason why there are more than 5800 unit tests in the -testEidos suite, and why those tests are hand-written by me rather than done in some more automated fashion.  Testing languages is really hard.  That said, I think in *principle* automated cross-testing between R and Eidos could be a very useful thing to do, and this looks like a reasonable first run at it (but I haven't looked at it closely yet, since I wanted to have this discussion first).  This sort of thing is why I think we really desperately need a full-time QA engineer.  Even if this PR is a good idea and a good first step – as it probably is – I won't have time to do anything with it, and I suspect you won't have much either.  Where does that leave us?

Sorry if these comments feel negative.  I truly don't mean it that way – I would love love love to see something like this that would really comprehensively exercise Eidos beyond my unit tests.  I just found a whole boatload of bugs related to NAN in floating-point calculations in Eidos, in many different functions, that slipped through the cracks because the unit tests I wrote didn't cover all the cases, and I didn't fully understand the subtleties of how NANs work in some types of calculations.  I've got a commit on my machine to fix all that; and I don't doubt there's more where that came from (but it's finding the bugs that's the hard part! :->).  So I'm totally on board with the idea here.  I guess I just want to make it clear how difficult the problem is here that we need to solve, and ask: who will actually have the time/energy to take this where it needs to go?  Testing that, say, sin() of a vector of random numbers produces the same result in both languages is pretty easy – but is highly unlikely to be where the interesting bugs lie.  Comprehensively testing every code path in Eidos is much, much harder.

petrelharp:
Fair enough! Although: those things sound pretty straightforward to me to deal with. I mean, it'd be a fair bit of work, sure, but notice that for instance I've already got it set up to [test on vectors of various lengths and types](https://github.com/MesserLab/SLiM/pull/104/files#diff-530238f72c603dd7b93cd2ac0f22905eR74). But you're right - I shouldn't be writing this whole testing framework from scratch right now, as satisfying as it would be. Let's file this away, but feel free to bug me about it in the future.

bhaller:
If we ever do get a hire for QA, this should probably be one of the first things we have that person work on.

Issue #100, opened 2020-07-22
add `quantile` to Eidos


mufernando:
In #74 I inquired about adding a `quantile` function to Eidos. I bet this would have many uses, but most importantly it is essential for simulating truncation selection in SLiM.

@bhaller rightfully raised the issue that `R`s implementation of `quantile` includes many different methods, which differ in how they handle interpolation. I think we don't need anything fancy, so I propose we implement a mid-point interpolation. That is, when  a quantile lies between two data points `i<j`, we return `(i+j)/2`. This is really equivalent to [`numpy`s midpoint interpolation](https://numpy.org/devdocs/reference/generated/numpy.quantile.html).

To get started, I wrote a Eidos `quantile` function that does exactly what I said above:

```
function (numeric)quantile(numeric x, float q) {
    pos = q*(length(x)-1);
    //midpoint impl
    idx = order(x);
    x_below = x[idx[floor(pos)]];
    x_above = x[idx[ceil(pos)]];
    return (x_below+x_above)/2;
}
```

@bhaller how does this sound? I can get some example arrays, quantiles and expected results for you to use as a test. It would be great if you could implement it, as I'm kind of swamped right now. But I don't think this is urgent.

mufernando:
ah, see here for some how numpy tests things:

https://github.com/numpy/numpy/blob/a39e3021b9304fb5a76542d444b7fec2dcff1374/numpy/lib/tests/test_function_base.py#L2697-L2703

bhaller:
Great, thanks for the Eidos implementation and the test, this is now very clear.  @petrelharp, does this seem to you to be the correct approach to take?  I ask since this is a math-y thing.  ;->  Note that I can always add a `type` optional argument to the function later on, preserving backward compatibility, if other types of quantile calculation are needed.

petrelharp:
Hm, good question. My strong prior is that we should do it however R does it, and probably the default method. However, `?quantile` says that R's choice is inherited from S, and maybe just locked in by history. 

Here's a comparison of the 9 methods:
```
n <- 6
x <- runif(n)
y <- rexp(n)
p <- seq(0, 1, length.out=101)
qx <- sapply(1:9, function (k) quantile(x, probs=p, type=k))
qy <- sapply(1:9, function (k) quantile(y, probs=p, type=k))

layout(1:2)
matplot(p, qx, type='l', lty=1, col=1:9)
points((1:n)/(n+1), sort(x))
[...]

PR #105, opened 2020-08-01
quantile fix?

petrelharp:
This is a more direct translation of the R code in #100.


petrelharp:
... and, it seems to do the right thing with INF, at least.

bhaller:
Passes all unit tests (I had a bunch written that were commented out, since I wanted to limit the spewage :->).  I've incorporated it into the diffs on my machine, since I'm working in the same file etc.; so I'm going to close this PR but I've really merged it.  :->

PR #101, opened 2020-07-29
update tskit

petrelharp:
Changes (see the [tskit changelog](https://github.com/tskit-dev/tskit/blob/master/c/CHANGELOG.rst)):

- There is now a mutation.time column! So, `tsk_mutation_table_X` methods need to use it.
- table collections loaded from files no longer live on top of the underlying kastore memory.
- vargen genotypes are now ints, not uints.
- We can now disable table collection integrity checking when sorting tables, which we do when not in a DEBUG build.
- Everything now has metadata, so `tsk_edge_table_X` and `tsk_migration_table_X` now have metadata and metadata length arguments. But, we don't use these at all, so this is trivial.

TODO:

- <s>Remove mutation time from the mutation metadata, and use the mutation time column instead. (I think we should do this? This is a file format change!)</s> (not doing this)
- Set the metadata schemas on the tables.

This hasn't been tested at all yet. It would be nice to have some developer documentation that says how to run the tests.

PR #106, opened 2020-08-28
remove first gen

petrelharp:
Here's a start at this.

Also, adding "stage" to the metadata.

I don't *think* we need to do anything different on reading in files: these may have individuals marked FIRST_GEN, but these are not samples, so everything should just work out. (They may persist through luck, and keep the flag, but that's ok.)  

Simliarly, we don't set the stage on loading the tree sequence - although we could check for consistency.

This has all the changes in #101 as well, so the diff will look reasonable once that's merged.


bhaller:
Hey.  The previous PR is merged now, but the "files changed" for this PR still seems to include all the diffs from that previous PR.  Maybe this PR needs to be rebased or something?  Anyway, I'll wait for you to fix this situation before I look at it further.  :->

petrelharp:
hm, yep, needed a rebase for some reason.

core/slim_globals.h
```diff
@@ -352,6 +352,7 @@ extern EidosValue_String_SP gStaticEidosValue_StringT;
 	
 #define SLIM_TSK_INDIVIDUAL_ALIVE       ((uint32_t)(1 << 16))
 #define SLIM_TSK_INDIVIDUAL_REMEMBERED  ((uint32_t)(1 << 17))
+// DEPRECATED:
 #define SLIM_TSK_INDIVIDUAL_FIRST_GEN   ((uint32_t)(1 << 18))
```
bhaller, line 356: If we're no longer using this anywhere in the code, can we actually comment this line out, to ensure that nobody uses it?

core/slim_sim.cpp
```diff
@@ -6284,8 +6229,25 @@ void SLiMSim::WriteTreeSequenceMetadata(tsk_table_collection_t *p_tables)
...
     // But that's being a headache, so we're skipping it.
     nlohmann::json metadata;
-	metadata["SLiM"]["model_type"] = (ModelType() == SLiMModelType::kModelTypeWF) ? "WF" : "nonWF";
-	metadata["SLiM"]["generation"] = Generation();
+    if (ModelType() == SLiMModelType::kModelTypeWF) {
+        metadata["SLiM"]["model_type"] = "WF";
+        if (GenerationStage() == SLiMGenerationStage::kWFStage1ExecuteEarlyScripts) {
+            metadata["SLiM"]["stage"] = "early";
+        } else {
+            assert(GenerationStage() == SLiMGenerationStage::kWFStage5ExecuteLateScripts);
```
bhaller, line 6237: I guess this assert is OK-ish.  I don't think we prohibit people from calling treeSeqOutput() from other places – a reproduction() callback, say.  But maybe we should?  Maybe add that to our list of things we should tighten up to make our lives easier?

core/slim_globals.h
petrelharp: Yes, I guess so - I thought we'd need to keep it around to load old files, but since we don't...

core/slim_sim.cpp
```diff
@@ -6435,7 +6397,7 @@ void SLiMSim::WriteProvenanceTable(tsk_table_collection_t *p_tables, bool p_use_
 	j["metadata"]["individuals"]["flags"]["17"]["name"] = "SLIM_TSK_INDIVIDUAL_REMEMBERED";
 	j["metadata"]["individuals"]["flags"]["17"]["description"] = "the individual was requested by the user to be remembered";
 	j["metadata"]["individuals"]["flags"]["18"]["name"] = "SLIM_TSK_INDIVIDUAL_FIRST_GEN";
-	j["metadata"]["individuals"]["flags"]["18"]["description"] = "the individual was in the first generation of a new population";
+	j["metadata"]["individuals"]["flags"]["18"]["description"] = "the individual was in the first generation of a new population (DEPRECATED)";
```
bhaller, line 6400: Since we no longer write this flag out (right?), do we need to describe it in our metadata?

core/subpopulation.cpp
```diff
@@ -730,10 +730,6 @@ void Subpopulation::GenerateParentsToFit(slim_age_t p_initial_age, double p_sex_
 			parent_individuals_.push_back(individual);
 		}
 	}
-	if (recording_tree_sequence)
-	{
-		sim.AddIndividualsToTable(parent_individuals_.data(), parent_individuals_.size(), nullptr, SLIM_TSK_INDIVIDUAL_FIRST_GEN);
```
bhaller, line 735: OK, so... how do individual table entries get added now?  Do we just say that there are no individual table entries for the first gen unless the user calls treeSeqRememberIndividuals() on them?

bhaller:
I just made a few comments.  Seems pretty straightforward, but I haven't thought about it very hard.  Does it pass all of your tests?  I can test it tomorrow with crosscheck, etc.

core/slim_sim.cpp
petrelharp: Oh, right, good point. I vote to tighten that up - dealing with that correctly in pyslim will be a pain. The error could even say "if you really want to do this please file a github issue" - but, I can't currently imagine a situation where someone would actually need to do this.

core/subpopulation.cpp
petrelharp: yes!

petrelharp:
I haven't tested yet - I am still getting the pyslim changes together. I'll let you know when that all works.

core/slim_sim.cpp
```diff
@@ -6052,33 +6048,26 @@ void SLiMSim::AddIndividualsToTable(Individual * const *p_individual, size_t p_n
...
+            auto ind_pos = std::find(remembered_individuals.begin(), remembered_individuals.end(), ped_id);
+            
+            if (ind_pos == remembered_individuals.end())
+                tsk_individual = TSK_NULL;	// not in the table already
+            else
+                tsk_individual = (tsk_id_t)std::distance(remembered_individuals.begin(), ind_pos);
+        }
+        else
+        {
+            // this case has a fash search (hash table), but the unordered_map is slow to build
```
bhaller, line 6063: might as well fix the typo while you're modifying this code, fash->fast

core/slim_sim.cpp
```diff
@@ -6148,50 +6137,6 @@ void SLiMSim::AddCurrentGenerationToIndividuals(tsk_table_collection_t *p_tables
 	}
 }
 
-void SLiMSim::UnmarkFirstGenerationSamples(tsk_table_collection_t *p_tables)
```
bhaller, line 6151: oh man, this feels good :->

treerec/tskit/tables.c
```diff
@@ -6458,7 +6458,7 @@ simplifier_insert_input_roots(simplifier_t *self)
                     if (ret != 0) {
                         goto out;
                     }
-                    simplifier_map_mutations(self, input_id, x->left, x->right, x->node);
+                    simplifier_map_mutations(self, input_id, x->left, x->right, output_id);
```
bhaller, line 6461: Sounds like this change will be in a tskit point-release soon; I guess we're taking it early, and will take the rest of the tskit changes that are in that point-release with a separate PR, also for SLiM 3.5, right?  And then hopefully we're done with all this tskit fallout for a while.

core/slim_sim.cpp
petrelharp: oh I thot u meant "fashionable search" ;)

Issue #116, opened 2020-11-07
Invalid filname for windows filesystems: `QtSLiM/recipes/Recipe 17.4 ...`


grahamgower:
It seems that windows doesn't allow filenames containing double quotes.
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

Attempting to clone the repository on windows produces the following:
```
Cloning into 'SLiM'...
error: invalid path 'QtSLiM/recipes/Recipe 17.4 - Detecting the "dip in diversity" (analyzing tree heights in Python) I.txt'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

bhaller:
I'll fix this, since of course it's easy to do so, but Windows is not a supported platform.  Since people report no problems checking out SLiM under the WSL, it looks like the GH Action you're setting up is doing the checkout of the sources outside of the WSL.  But anyway, easy to fix so I'll fix it, and maybe the rest of the GH Action will happen inside the WSL even if the checkout doesn't?

grahamgower:
Ah yes, it looks like there are some hoops to jump through to enable and use WSL on github actions. I'm not suggesting you should spend your energy supporting Windows, but not being able to clone the repository would be a significant barrier to any vollunteer that wanted to try. At the very least, the github action could be configured to do non-build things for windows to catch any such recurrence of this. It's up to you though!

bhaller:
Yes, makes sense.  :->

PR #117, opened 2020-11-07
github action tests

grahamgower:


bhaller:
Hi @grahamgower.  GH says this is "still a work in progress" and can't be merged.  What is it?  :->

grahamgower:
It's a continuous integration config file, like the one for travis, but this one is for github actions. I think github actions need to be enabled for the SLiM repository (somewhere under the repo settings), but you can see what works and doesn't work on the branch under my fork (https://github.com/grahamgower/SLiM/actions/runs/351263364).

bhaller:
"Allow all actions" was already chosen in SLiM's settings.  (Should it be "local actions only" for security?)

Interesting that they have a macOS 11.0 build.  Also interesting that it fails.  :->

Windows isn't supported by SLiM, except for in Linux running under the WSL, so no surprise that those tests are failing.  I'd say you could comment them out for the time being, since there are presently no plans to support Windows.

Seems groovy.  I assume that adopting this won't break my existing Travis CI, right?  It'll just add another CI option?

Thanks a bunch!  What remains to be done with it for me to accept it?

grahamgower:
> "Allow all actions" was already chosen in SLiM's settings. (Should it be "local actions only" for security?)

Oh, strange that it doesn't just run here then. I suspect that the "local actions only" option will prevent pull requests from running actions (but there's no more info in the docs), so probably the the default "allow all actions" is the right choice.

> Interesting that they have a macOS 11.0 build. Also interesting that it fails. :->

Yeah, it seems that they removed a file which the build system checks for. I guess the problem is in the toolchain, rather than something fixable in SLiM's cmake files?

> Windows isn't supported by SLiM, except for in Linux running under the WSL, so no surprise that those tests are failing. I'd say you could comment them out for the time being, since there are presently no plans to support Windows.

I think this is running under some posixy subsystem (I've configured it to use the bash shell, at least). So I was hoping it might build ok, but the repository can't even be cloned because of #116. So I don't think it will even be possible to build with WSL right now.

> Seems groovy. I assume that adopting this won't break my existing Travis CI, right? It'll just add another CI option?

Yep.

> Thanks a bunch! What remains to be done with it for me to accept it?

It's possible to select a variety of platforms, compilers, etc. with github actions. And the qt installer here can install almost any qt version.  So, do you have an idea of what versions of things you'd like to test? Probably the oldest version supported, plus the newest versions available? And maybe we could choose a really old gcc and a really new one.

I'll change the PR from "draft" to "ready for review" once I've played with the Windows build a bit (so, once #116 is fixed). And maybe I'll just remove macOS-11.0.

grahamgower:
> > "Allow all actions" was already chosen in SLiM's settings. (Should it be "local actions only" for security?)
> 
> Oh, strange that it doesn't just run here then.

I suspect this is not running under this repository (and only running on my fork) due to the security model github imposes---probably only a user with write access would be able to introduce a new github action and have it run under the repository before being merged. But I couldn't find this documented anywhere. In any event, once this PR gets merged the action should run against new PRs as expected (just like travis does now).

bhaller:
> > Interesting that they have a macOS 11.0 build. Also interesting that it fails. :->
> 
> Yeah, it seems that they removed a file which the build system checks for. I guess the problem is in the toolchain, rather than something fixable in SLiM's cmake files?

I'm not sure, I'll need to look into it.

> > Windows isn't supported by SLiM, except for in Linux running under the WSL, so no surprise that those tests are failing. I'd say you could comment them out for the time being, since there are presently no plans to support Windows.
> 
> I think this is running under some posixy subsystem (I've configured it to use the bash shell, at least). So I was hoping it might build ok, but the repository can't even be cloned because of #116. So I don't think it will even be possible to build with WSL right now.

I've fixed #116 on my machine now, will push it to GitHub when the other changes on my machine can go as well, which hopefully will be later today or tomorrow.

> > Thanks a bunch! What remains to be done with it for me to accept it?
> 
> It's possible to select a variety of platforms, compilers, etc. with github actions. And the qt installer here can install almost any qt version. So, do you have an idea of what versions of things you'd like to test? Probably the oldest version supported, plus the newest versions available? And maybe we could choose a really old gcc and a really new one.
> 
> I'll change the PR from "draft" to "ready for review" once I've played with the Windows build a bit (so, once #116 is fixed). And maybe I'll just remove macOS-11.0.

OK, this turned out to be pretty complicated to try to answer comprehensively.  Sorry, ugh.

### OS versions:

Ubuntu 18.04 and 20.04 are good, no need to go back further I don't think; nobody has ever asked for 16.04 support.

macOS 10.15 and 11.0; also macOS 10.10 if they go back that far, that's the minimum supported by SLiM for now.  If they don't have 10.10, then just whatever the oldest one post-10.10 is.  I'd keep the macOS-11.0 build in; I will need to figure that issue out sooner or later, and I think it may be something I can fix on my end given a little elbow grease.  But as I note below, on macOS-11.0 there's no need to try building QtSLiM, that will definitely not work.

OK, onward to the details.

**Building QtSLiM:**

This has complex versioning requirements since Qt has complex versioning requirements:

Qt versions: Qt 5.9.5 is the rock bottom minimum supported, because it is what Ubuntu 18.04 has preinstalled out of the box (although oddly Qt itself says that 5.9.x does not support Ubuntu 18.04!).  I suppose the other good ones to test against would be the LTS releases: 5.12 LTS (5.12.8 at present) and 5.15 LTS (5.15.1 at present, I believe).  If they offer any Qt 6 prerelease, don't bother with that, that's unsupported for the foreseeable future.

However, those Qt versions only work against specific versions of the OSes, so there's no need for a full grid, I don't think.  Let's do these combos:

Qt 5.9.5 with macOS 10.10 and Ubuntu 18.04.  This will test our oldest officially _supported_ version combos.

Qt 5.12 LTS (5.12.8 at present) with macOS 10.12 (since this Qt version only supports 10.12-10.14) and Ubuntu 18.04.  This will test our oldest _recommended_ version combo.

5.15 LTS (5.15.1 at present) with macOS 10.15 and Ubuntu 18.04 and 20.04.  This will test the _cutting edge_ combo.  Let's do it with both Ubuntu 18.04 and 20.04, though, since 20.04 is still not officially supported by Qt 5.15, apparently.

Oh, and don't bother trying to build QtSLiM on macOS 11.0 at all; Qt definitely doesn't support it yet.

**Building the `slim` and `eidos` command-line tools:**

This is less complicated since they don't depend on Qt.  They should be built, and then tested by running -testSLiM and -testEidos if possible, on macOS 10.10, 10.15, and 11.0, and on Ubuntu 18.04 and 20.04, period.

### Compiler/tool versions

Well, I don't think we need to do a comprehensive grid with compiler, cmake, etc.; that would just waste computing resources.  However, it would be good to test a few key combos.  So let's see.

**For building `slim` and `eidos`:**

The minimum cmake supported is 2.8.  The build can always just use whatever version of cmake is installed by the OS version we're using, unless it is older than 2.8; no need to test more recent cmake versions against old OSes.  The command-line tools don't build using qmake so that is irrelevant.  On macOS the compiler used should be whatever is the default installed by the OS (some version of clang).  On Ubuntu it would be good to test:

GCC 4.8.5 with 18.04; we require C++11 so that's probably as old as we can reasonably go
GCC 7.4.0 with 18.04; this is the default installed for Ubuntu 18.04
GCC 9.3 with 20.04; this is the default installed for Ubuntu 20.04
GCC 10.2 with 20.04; this is the current bleeding edge and may not work, I have no idea

That way we test old/old and new/new combos.

**For building QtSLiM:**

The minimum cmake supported is 3.1.0, which I think means it will need to be updated for some of the older platforms.  We can just always use whatever the installed cmake is unless it is older than 3.1.0, in which case go with 3.1.0; no need to test newer cmake versions, we've never had any issue there.  The qmake version should always be the one installed with the Qt version being used.  It would be good to test both the cmake and qmake builds on all platforms.  Compiler versions:

On macOS, use the default installed by the OS (clang something)
On Ubuntu 18.04, use whatever is installed, which I believe is GCC 7.4.0; we're not trying to support building QtSLiM with old compilers; could also try GCC 9.3 if possible.
On Ubuntu 20.04, use whatever is installed (GCC 9.3); Qt doesn't claim to support GCC 10.

Wow, that was really complicated, sorry.  I hope you can make sense of all the gibberish above!  Thanks again for doing this; it's appreciated, and will help make SLiM more robust.

bhaller:
Oh, and if the Windows builds still don't work after I push my fix for 116 later, I wouldn't spend much time on that.  I think the WSL is supposed to be a pretty good mimic of Linux; if it builds on the Ubuntu variants we include, I think it ought to build under the WSL too (assuming the user has correctly configured things), so I'm not too worried about that.

grahamgower:
Wow! thanks for the detailed info. There's only one macOS 10.x version available, unfortunately.
https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
But I'll keep poking things this week, and ping you when I think it's good shape.

grahamgower:
Hey @bhaller, I think this is looking pretty good now. I've broken things up into CLI builds/tests and GUI builds, so that CLI and GUI builds run concurrently (which hopefully makes things complete slightly quicker). If the GUI builds with Qt 5.9.5 and 5.15.1, it seems fairly unlikely that you'll experience problems with intermediate versions (this more likely indicates an upstream issue). Similarly, if SLiM builds with GCC 4.8 and 10.2, then the chance of problems with intermediate versions is vanishingly small. GCC 10.x includes some nice additional warnings which can sometimes catch bugs, so I reckon it's useful to build with that.

It looks like this has uncovered one minor issue with the qmake build for gcc 4.8. At a guess, I'd say the file should be getting built with a C compiler, not a C++ compiler, because `restrict` is a C-only keyword (so this might need a tweak to the qmake script(s)). I suppose the GNU folks just silently ignore this keyword for more recent g++ versions? Or maybe it got added to the C++ spec?
```
g++ -o core.o ../../../treerec/tskit/core.c
56
../../../treerec/tskit/core.c:619:42: error: expected ‘;’, ‘,’ or ‘)’ before ‘array’
57
 tsk_search_sorted(const double *restrict array, size_t size, double value)
```

grahamgower:
> the file should be getting built with a C compiler, not a C++ compiler, because `restrict` is a C-only keyword

Oh, sorry, I misread the error message because the long line wrapped in an unfortunate location. That file is definitely being compiled with the `gcc`, not `g++`. So I'm not sure why it's complaining about the `restrict` keyword!

bhaller:
My guess is that --std=c99 needs to be specified.  I just added that to the qmake stuff and pushed; not sure if that triggers an automatic rebuild on your end yet, or if you need to rebase your PR or something.  I also fixed a whole bunch of warnings from the QtSLiM build that I don't see on my local machine, so that's useful (although none of them actually mattered).  Probably a whole bunch of warnings still left to fix, I'll do more passes as needed.

This all looks great.  Should we merge?

grahamgower:
> This all looks great. Should we merge?

Yes, please! That should trigger the actions in the SLiM repository, so we'll see if your change(s) do indeed fix that particular build.

bhaller:
OK, merged

grahamgower:
Great! The actions tab is now a hive of activity. Looks like there's a different error for the qmake/gcc-4.8 build now.

bhaller:
I guess I'll try to fix all of these errors and warnings on my end, and I'll ping you if there seems to be a problem with the "GH action" end of things; it seems like that end of things is working fine, and there are just some new build issues to be sorted out since these build parameters are a bit different than we've ever used before.  Thanks!

bhaller:
OK, I got that last build to work by banging on it with a hammer until it stopped complaining.  :->  All builds now work, and trigger automatically.  Everything seems happy.  Thank you thank you thank you!  This is great.  :->

bhaller:
Hi @grahamgower.  FYI, GitHub Actions is showing me a warning:

> Check warning on line 1 in .github
> 
> github-actions / canceller
> 
> .github#L1
> Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816

Not sure if this is of any importance.  :->  Maybe something needs to be tweaked so that it continues to build with Ubuntu 18.04 as intended, instead of changing to 20.04 unintentionally?

And by the way, maybe it would be good if I had contact info for you outside of GitHub, so this could be an email rather than a comment on a closed pull request...?  :->  How do you like to communicate about such things?

grahamgower:
Hey @bhaller, this is definitely benign. Github seem to be using 'github actions' warnings to communicate platform changes/updates to their users. In this case, the canceller job should run just fine on any OS. I'll submit a pr to pin the OS version anyway.

You're welcome to contact me here, on the tskit-dev slack, or by email.

bhaller:
Good to get rid of the warning in any case.  :->  Thanks!

PR #129, opened 2021-01-06
Change initializeTreeSequence to initializeTreeSeq

hyanwong:
I think that in the docs/etc sometimes the wrong function name is used. From my reading it should be `initializeTreeSeq()` (small "i", no "ence" at the end).


bhaller:
Yep, thanks!

PR #131, opened 2021-01-07
update to implementation to reflect PR#106 (closes #130)

### LINKED ISSUE #130: Update description of treerec implementation
The [implementation.md](https://github.com/MesserLab/SLiM/blob/master/treerec/implementation.md) file describes saving the first generation individuals for recapitation, but as of SLiM 3.5 I think this doesn't apply, so I guess this file needs updating? I'm updating it as part of https://github.com/MesserLab/SLiM/pull/40#issuecomment-755222254 but I thought I should check here that this is correct, and whether we want to keep the `implementation.md` file anyway.

petrelharp:
How's this, @hyanwong?

### DIFF
--- a/treerec/implementation.md
+++ b/treerec/implementation.md
@@ -5,14 +5,13 @@
 Both individuals and nodes (= genomes) have flags. We use individual flags for
 internal bookkeeping in SLiM; while msprime uses the SAMPLE flag of nodes.
 
-0. The first generation individuals get entered into the individual table, with
-    the `INDIVIDUAL_FIRST_GEN` flag, so we know to keep these around through simplify.
-    Their genomes' node IDs are also added to `remembered_genomes_`.
-
 1. Every new individuals' genomes each get the `NODE_IS_SAMPLE` flag,
     because the SAMPLE flag means that we have whole-genome information about a genome,
     which at this point we do. Node times are recorded as the number of generations
-    before the start of the simulation (i.e., -1 * sim.generation).
+    before the start of the simulation (i.e., -1 * sim.generation). However, the
+    individuals themselves are *not* added, in part because much of the information
+    in the individual table (e.g., location, age) may change, so the user should have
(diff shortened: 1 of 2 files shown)


treerec/implementation.md
```diff
@@ -5,10 +5,6 @@
 Both individuals and nodes (= genomes) have flags. We use individual flags for
 internal bookkeeping in SLiM; while msprime uses the SAMPLE flag of nodes.
 
-0. The first generation individuals get entered into the individual table, with
-    the `INDIVIDUAL_FIRST_GEN` flag, so we know to keep these around through simplify.
-    Their genomes' node IDs are also added to `remembered_genomes_`.
-
 1. Every new individuals' genomes each get the `NODE_IS_SAMPLE` flag,
```
hyanwong, line 8: This wording suggests that every individual is added, whereas I think that (at the moment) only "remembered" and "alive" are entered. Should we say "If an individual is added, it initially gets the `NODE_IS_SAMPLE` flag, ..."

hyanwong (commented):
All looks great. Just one suggestion, if I've understood correctly.

hyanwong:
I just notice that you can probably delete this sentence too:

https://github.com/MesserLab/SLiM/blob/35838d3cb0add18e216d36465e5ed64194a34cd6/core/slim_sim.cpp#L6120

bhaller:
@petrelharp, if I understand the comments above correctly, @hyanwong has proposed a change or two that is not yet present in this PR, so I'm holding off on merging.  Ping me when this is ready to merge, thanks.

petrelharp:
I think this is good to merge, @bhaller - I've addressed both of Yan's comments (and I've rebased it).

bhaller:
Thanks, merged.  GitHub is offering to let me delete the branch; not sure what the etiquette is on that, I'm surprised it would even let me given that it's your branch not mine!  :->

PR #135, opened 2021-01-10
Use new tskit param

hyanwong:
To avoid warning issued as of tskit 0.3: "FutureWarning: The impute_missing_data parameter was deprecated in 0.3.0 and will be removed. Use ``isolated_as_missing=False`` instead of``impute_missing_data=True``"

Do we need to also specify a minimum tskit version of (say) 0.3.4 somewhere @petrelharp ?


bhaller:
I think this was done by @petrelharp's PR just now, if I understand correctly?  So... how does one reject/close a PR?  Oh, I see a "close with comment" button below.  I assume you can re-open this if I have misunderstood.

PR #136, opened 2021-01-10
switch from nose to pytest, closes #134

### LINKED ISSUE #134: Switch to pytest instead of nose for python treeseq tests?
In `treerec/tests/do_test.sh` we (i.e. @petrelharp I presume) use the `nose` module for unit testing, which is no longer being actively maintained. Might it be useful to switch to `pytest`, like we have done in the various tskit projects? I think it might be as simple as `s/nose/pytest/`, but perhaps @benjeffery would know, and might be able to suggest ways to use this to parallelise the tests (perhaps they are parallelised already, and I'm misunderstanding here).

petrelharp:
This does the (easy) job of switching to pytest (and I also incorporated #135 before I noticed you'd done that). Clearly the how we run the tests could be improved (eg parallelization) but I'm not going to jump into that at the moment. Note that these are supplemental tests, which just check for consistency between the tree sequence and what SLiM says is going on, to @bhaller's much more comprehensive set of tests.

### DIFF
--- a/treerec/tests/do_test.sh
+++ b/treerec/tests/do_test.sh
@@ -33,14 +33,14 @@ do
     if [ -e "test_output/slim_no_mutation_output.txt" ]
     then
         echo "  Testing no-mutation output."
-        python3 -m nose test_no_mutation_output.py || exit 1
+        python3 -m pytest test_no_mutation_output.py || exit 1
         TESTED="yes"
     fi
 
     if [ -e "test_output/slim_mutation_output.txt" ]
     then
         echo "  Testing with-mutation output."
-        python3 -m nose test_mutation_output.py || exit 1
+        python3 -m pytest test_mutation_output.py || exit 1
         TESTED="yes"
     fi
 
--- a/treerec/tests/test_mutation_output.py
+++ b/treerec/tests/test_mutation_output.py
@@ -12,13 +12,13 @@ class TestWithMutations(TestSlimOutput):
         slim = {}
         for header in slim_file:
             headstring = header.split()
-            self.assertEqual(headstring[0], "#Genome:")
+            assert headstring[0] == "#Genome:"
             genome = int(headstring[1])
             mutations = slim_file.readline().split()
-            self.assertEqual(mutations[0], "Mutations:")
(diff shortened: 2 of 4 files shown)


bhaller:
Merged, thanks!

hyanwong:
Cool, thanks. I started thinking about parallelisation and replacing the `do_tests.sh` script with pytest fixtures, something like:

```
# in conftest.py
import os

recipe_dir = "testRecipes"

def pytest_addoption(parser):
    parser.addoption("--recipe", action="store", default=None, help="The name of the script to run in " + recipe_dir)

def pytest_generate_tests(metafunc):
    # This is called for every test. Only get/set command line arguments
    # if the argument is specified in the list of test "fixturenames".
    recipe = metafunc.config.option.recipe
    if recipe is None:
        # Test everything begining with "test_" in recipe_dir
        filelist = [d for d in os.listdir(recipe_dir) if d.startswith("test_")]
        metafunc.parametrize("recipe", filelist)
    else:
        metafunc.parametrize("recipe", [recipe])
```

And then saving separate test outputs in their own pytest `temp_dir`. But then I realised that it would mean re-running the SLiM simulations for each test class, and I wondered if there was a way to only run SLiM once each for each recipe in ` metafunc.parametrize`, so I stopped there. I'm sticking this comment down in case it helps if/when we revisit this.

petrelharp:
If you've got the appetite for things like this, I'd appreciate help over at https://github.com/tskit-dev/pyslim/issues/114 first... =)

hyanwong:
> If you've got the appetite for things like this, I'd appreciate help over at [tskit-dev/pyslim#114](https://github.com/tskit-dev/pyslim/issues/114) first... =)

I just had a quick look, but I must say I'm baffled at the setup. Running with `pytest -n 4` does indeed abort for me with 

```
    def get_slim_example(self, name, return_info=False):
        ex = example_files[name]
        treefile = ex['basename'] + ".trees"
        print("---->", treefile)
>       assert os.path.isfile(treefile)
E       AssertionError
```

But I don't know how you are creating the files in the first place. I guess it's quite convoluted. Care to explain how it works?

PR #142, opened 2021-01-14
Major reorganization of python tests

hyanwong:
Here's my suggestion for restructuring the pytest stuff in SLiM, @petrelharp 

I suspect it runs a bit slower, but it can be parallelised, so 👍 .

There's a lot of changes here. The overview is:

Each pytest session runs SLiM so that the working directory is a subdirectory of temp dir provided by pytest; the subdirectory is simply given the name of the test. That means any other thread trying to get the SLiM output will be locked until the files in that subdirectories are produced (once per pytest session). To get at the slim setup files (which were previously called "testing_utils.slim" and "marked_mutations_setup.slim", which I renamed to "init.slim" and "init_marked_mutations.slim"), they are placed in a directory called `setup_slim`, and symlink to that dir is generated from within the pytest temp dir. A bit hacky, but it works for me, and I couldn't work out how to pass absolute paths into SLiM for the `source()` function.

I hacked some of the special slim code that initialises the simulations to make the function naming more consistent, and so that every time an `outputXXX` function is called, SLiM saves a new set of files to a subdirectory of the current SLiM working directory, with the subdirectory named after the generation that the output happened. This should allow us to e.g. output files both in the middle of a simulation and at the end (although I haven't actually tried this yet). So for instance, if we call `outputIndividuals()` in generation 100 from the script `test_052_retain_individuals_nonWF.slim`, the various `.trees` and output files will be stored in `pytest_tmp_dir/test_052_retain_individuals_nonWF/100/*`. In thew same simulation we could the also output in generation 200, and another (different) set of files would be saved in `pytest_tmp_dir/test_052_retain_individuals_nonWF/100/*`. There's some hackery to allow data output in previous generations (via `addAncestralSamples`) to be included in the files saved at the end of the sim (or whenever the outputXXX function is called), but it should all work.

For simplicity, I've put all the test functions in the same file, and removed the utils module you made, rolling the global functions into the top of the file: I hope that's OK. To determine whether a test (e.g. for consistency of individuals) should be run, the test looks for the file called (in this case) `slim_individual_output.txt`, and if it exists, runs the test for individual consistency with the TS. If the file doesn't exist, the test is bypassed. That's a bit of a hack - you might be able to think of a better way to do it.

The SLiM recipes are passed as parameters to a pytest "fixture" called "session_data", which returns a list of the directories in which data is found (e.g. `test_052_retain_individuals_nonWF/100`, `test_052_retain_individuals_nonWF/200`, etc). That means no more bash files (hurrah). All the tests can be run on a specific recipe by e.g. `pytest -k test_052_retain_individuals_nonWF`. Or a subset of tests can be run by e.g. `pytest test_consistency.py::TestIndividuals` (in this case, the TestIndividuals tests will be run on all the scripts, even though only a few actually output files that allow individuals to be tested, so it's a bit wasteful, but who cares?)

A separate test checks that all the scripts output at least one XXX_output.txt file in each output subdirectory. Again, this seems like a bit of a hack, but it was the simplest way I could think to check that at least something is output by the script.

This is all quite convoluted, and I don't know if it will port very well to the test framework in pyslim. But I thought it was worth a try to see if we could "do it right".


petrelharp:
*NICE!*  And, it works! And, I understand fixtures now! (The pictures of physical test fixtures at the bottom of the Wikipedia article really helped, for some reason.)

Let's see: 

> To get at the slim setup files (which were previously called "testing_utils.slim" and "marked_mutations_setup.slim", which I renamed to "init.slim" and "init_marked_mutations.slim"), they are placed in a directory called setup_slim, and symlink to that dir is generated from within the pytest temp dir. 

Instead, how about we just run SLiM from the current directory, instead of changing directories, but invoke it like `slim -d OUTDIR="tmpdirname" -s 23 recipe.slim`, and insert `OUTDIR` before all the output files? It looks like we'd just need to change e.g. [these lines](https://github.com/MesserLab/SLiM/pull/142/files) to
```
defineGlobal("mutationsFile", OUTDIR + "/slim_mutation_output.txt");
defineGlobal("individualsFile", OUTDIR + "/slim_individual_output.txt");
```
and calls to `setwd` like [this one](https://github.com/MesserLab/SLiM/pull/142/files#diff-ce2af117387304df808ba6698b059ce28e46c532006dec25fce8caf9dfcef74fR138) would instead modify `OUTDIR`? Maybe there's another global variable named `SAVEDIR` that gets changed, so we output to `OUTDIR/SAVEDIR` and  [`setwdForGen`](https://github.com/MesserLab/SLiM/pull/142/files#diff-ce2af117387304df808ba6698b059ce28e46c532006dec25fce8caf9dfcef74fR123) just does
```
defineGlobal("SAVEDIR", paste(sim.generation));
```

> To determine whether a test (e.g. for consistency of individuals) should be run, the test looks for the file called (in this case) slim_individual_output.txt, and if it exists, runs the test for individual consistency with the TS.

I agree, this feels fragile. Instead, how about we do this the [way it's done in pyslim](https://github.com/tskit-dev/pyslim/blob/main/tests/__init__.py#L28): so, [`slim_recipes`](https://github.com/MesserLab/SLiM/pull/142/files#diff-c3cd856a42b40422a770017ff7847952d155a45afe978c3f4adb55b1e3dea309R11) would not be an automatically-generated list of recipes, but instead be manually entered with information about which tests to run on that recipe. So, adding a new test we'd need to not just add the `.slim` file to `testRecipes/`, but also add its information to this location - but, we can specify which tests we expect it to run, which seems important. For instance, we'd have
```
slim_recipes = [
   {"file" : "test_000_ancestral_marks.slim", "mutation_output" : True},
   # some more files
]
```
and then the [`session_data` fixture](https://github.com/MesserLab/SLiM/pull/142/files#diff-c3cd856a42b40422a770017ff7847952d155a45afe978c3f4adb55b1e3dea309R53) would do like
```
    out = request.param
    out["data_dirs"] = data_dirs
    return out
```
in `TestWithMutations` instead of [skipping it if `output_file` isn't there](https://github.com/MesserLab/SLiM/pull/142/files#diff-c3cd856a42b40422a770017ff7847952d155a45afe978c3f4adb55b1e3dea309R181) we do
```
if "mutation_output" in session_data:
    for subdir in session_data["data_dirs"]:
         data_file = os.path.join(subdir, self.output_file)
        assert os.path.isfile(data_file)
```
(Note that the value of `session_data["mutation_output"]` isn't used for anything - it's just presence that's used - but we could stick information in there later if we wanted.)

What do you think?

hyanwong:
> Instead, how about we just run SLiM from the current directory, instead of changing directories, but invoke it like `slim -d OUTDIR="tmpdirname" -s 23 recipe.slim`, and insert `OUTDIR` before all the output files?

Yes, whatever you think easier. I was trying to avoid specifying stuff in SLiM because (a) I don't know it very well and (b) it's nice to be able to run the SLiM script by itself, without the test suite (and if we specified OUTDIR, that would have to exist otherwise the script would die). But if you prefer specifying in SLiM, then we can fix (b) in SLiM world by defaulting OUTDIR to `.` if it's not defined on the command line?

> It looks like we'd just need to change e.g. [these lines](https://github.com/MesserLab/SLiM/pull/142/files) to
> 
> ```
> defineGlobal("mutationsFile", OUTDIR + "/slim_mutation_output.txt");
> defineGlobal("individualsFile", OUTDIR + "/slim_individual_output.txt");
> ```

Yes, and the one that redefines the mutationsFile when there are no mutations. By the way, I don't understand why we do nothing when adding AncestralSamples when we are tracking marked mutations, but add all the data to the output file when we are putting SLiM mutations in. I wondered if that simply hadn't been coded up (hence my FIXME remark on line 108 of `init.slim`)

> I agree, this feels fragile. Instead, how about we do this the [way it's done in pyslim](https://github.com/tskit-dev/pyslim/blob/main/tests/__init__.py#L28): so, [`slim_recipes`](https://github.com/MesserLab/SLiM/pull/142/files#diff-c3cd856a42b40422a770017ff7847952d155a45afe978c3f4adb55b1e3dea309R11) would not be an automatically-generated list of recipes, but instead be manually entered with information about which tests to run on that recipe. So, adding a new test we'd need to not just add the `.slim` file to `testRecipes/`, but also add its information to this location - but, we can specify which tests we expect it to run, which seems important. For instance, we'd have

I'm a little hesitant to have an external file, because I never remember to keep the spec file in sync with the tests. And it's nice that if we add another recipe, it automatically gets tested, without having to edit a settings file. Also, in this case it also seems like there's redundancy in there: if we call outputXXX, we know we've done so because the XXX file is produced, so can be tested.

I wonder if we should somehow specify the tests within the SLiM simulation somehow? That would mean only editing one file (the recipe), and we could add something in e.g. `outputMutationResults()` that added all the tests in the class `TestWithMutations` to the list of tests-to-be-run. But this is all rather complicated, so perhaps your method of a config file is easier to understand? You'll know best what suits the `pyslim` test suite (talking of which, do we think this is the right idea for those tests too?)

hyanwong:
> Yes, whatever you think easier. I was trying to avoid specifying stuff in SLiM because (a) I don't know it very well and (b) it's nice to be able to run the SLiM script by itself, without the test suite (and if we specified OUTDIR, that would have to exist otherwise the script would die). But if you prefer specifying in SLiM, then we can fix (b) in SLiM world by defaulting OUTDIR to `.` if it's not defined on the command line?
> 
> > It looks like we'd just need to change e.g. [these lines](https://github.com/MesserLab/SLiM/pull/142/files) to
> > ```
> > defineGlobal("mutationsFile", OUTDIR + "/slim_mutation_output.txt");
> > defineGlobal("individualsFile", OUTDIR + "/slim_individual_output.txt");
> > ```

Oh, and I couldn't find a method in SLiM to join paths together, so I wasn't sure if specifying a path in a SLiM global like this would break on OSes where the path separator is (say) `:` or `\`. Dunno if there are many systems like that we need to worry about, though?

petrelharp:
>  if you prefer specifying in SLiM, then we can fix (b) in SLiM world by defaulting OUTDIR to . if it's not defined on the command line?

I usually do this by doing
```
if (!exists("OUTDIR")) defineGlobal("OUTDIR", ".");
```

petrelharp:
>  I don't understand why we do nothing when adding AncestralSamples when we are tracking marked mutations, but add all the data to the output file when we are putting SLiM mutations in. I wondered if that simply hadn't been coded up (hence my FIXME remark on line 108 of init.slim)

Oh, gee. I don't remember, and looking at the code hasn't jogged my memory yet. This looks a bit suspect, since we're maybe not actually remembering individuals when we think we are?

bhaller:
> Oh, and I couldn't find a method in SLiM to join paths together, so I wasn't sure if specifying a path in a SLiM global like this would break on OSes where the path separator is (say) `:` or `\`. Dunno if there are many systems like that we need to worry about, though?

SLiM only runs on Unix-based OSes; macOS is Unix-based, of course, and even when it's running on Windows it's running under the WSL (I don't think there's much interest in getting it to run natively on Windows).  So the path separator is / everywhere, at least for the foreseeable future.

petrelharp:
> I'm a little hesitant to have an external file, because I never remember to keep the spec file in sync with the tests. And it's nice that if we add another recipe, it automatically gets tested, without having to edit a settings file. Also, in this case it also seems like there's redundancy in there: if we call outputXXX, we know we've done so because the XXX file is produced, so can be tested.

Ok, but how do we say "we expect this recipe to run this test"? It'd be pretty easy to add a recipe and *think* that it will automatically run some test, but if it fails to output a file in the right place, right now it'll silently just not run.

hyanwong:
> SLiM only runs on Unix-based OSes; macOS is Unix-based, of course, and even when it's running on Windows it's running under the WSL (I don't think there's much interest in getting it to run natively on Windows). So the path separator is / everywhere, at least for the foreseeable future.

Thanks @bhaller  - I guessed it would be OK, but I was a little circumspect about hard coding it like this without any guarantees. Also, don the SLiM reading/writing routines cope OK with e.g. spaces in the path names? I assume so.

bhaller:
> > SLiM only runs on Unix-based OSes; macOS is Unix-based, of course, and even when it's running on Windows it's running under the WSL (I don't think there's much interest in getting it to run natively on Windows). So the path separator is / everywhere, at least for the foreseeable future.
> 
> Thanks @bhaller - I guessed it would be OK, but I was a little circumspect about hard coding it like this without any guarantees. Also, don the SLiM reading/writing routines cope OK with e.g. spaces in the path names? I assume so.

They should work fine.  If you guys end up needing anything from Eidos/SLiM for all this that isn't presently there (or doesn't presently work right!), just let me know.

hyanwong:
I think this is now ready to merge. I've refactored it along the lines of the pyslim tests, and added a recipe_specs.py file to flag up which tests to run on which recipes, just like in pyslim. Instead of creating symlinks, I realised the easiest thing to do was to load up the extra slim files at the start, then `setwd` at the very end of the sourced files,  to the temp dir we expect to use to save files. I've now stuck most of the Python logic in `conftest.py`, which should allow us to create new test files easily. I've also changed the recipes directory to `test_recipes` rather than `testRecipes`, to match pyslim.

It would be good to merge this, then I can get on with creating some tests for keep_unary_in_individuals.

treerec/tests/testing_utils.slim
```diff
@@ -1,4 +1,4 @@
-// Note call to initializeTreeSeq at the end.
+""// Define various functions used in SLiM testing
```
petrelharp, line 1: spurious `""` here?

hyanwong: Oops. Sorry

petrelharp:
LGTM! So much better than what we had before.  Hm - we should add these to `.github/workflows/tests.yml` so they're automatically run.

treerec/tests/testing_utils.slim
hyanwong: That file isn't used - it's an old named version that I forgot to delete. Will delete it and update.

hyanwong:
> we should add these to `.github/workflows/tests.yml` so they're automatically run.

Do you want to do this or shall I?

treerec/tests/testing_utils.slim
hyanwong: Done

petrelharp:
> > we should add these to `.github/workflows/tests.yml` so they're automatically run.
> 
> Do you want to do this or shall I?

Might as well do a new PR for that, in case it's not straightforward. I'd be happy to let you do it... I can't do it immediately, at least. Issue: https://github.com/MesserLab/SLiM/issues/151

petrelharp:
Ok, @bhaller, I think we can merge this now.

bhaller:
> Thanks for looking over it @petrelharp . I'm not sure why there are a number of `test_recipes` all starting with `test_000_`. I wasn't sure where the `000` came from (e.g. a section of the manual, or just an arbitrary number) so I didn't want to change any existing names, but perhaps I should?

@hyanwong should I merge or wait for this question to resolve?

hyanwong:
> @hyanwong should I merge or wait for this question to resolve?

Merge it now. I can always change the file names in a later PR, as I will be putting a few more tests in as part of retainCoalescentOnly.

bhaller:
OK, done.  Thanks!

petrelharp:
Sure, the `000` can be removed. The tests with numbers that aren't 000 were derived from the corresponding SLiM test recipes, so maybe that's useful information to keep around (but only for non-000 ones).

bhaller:
> Sure, the `000` can be removed. The tests with numbers that aren't 000 were derived from the corresponding SLiM test recipes, so maybe that's useful information to keep around (but only for non-000 ones).

I suppose if the tests are executed in alphabetical order the 000 causes those tests to execute first (?).  Is that desirable?

hyanwong:
> I suppose if the tests are executed in alphabetical order the 000 causes those tests to execute first (?). Is that desirable?

One of the points of redoing the tests was that they can all now be run in parallel, so I guess the order is (should be) irrelevant.

hyanwong:
> Sure, the `000` can be removed. The tests with numbers that aren't 000 were derived from the corresponding SLiM test recipes, so maybe that's useful information to keep around (but only for non-000 ones).

That's useful to know, thanks. I'll add it to the readme in my next PR.

bhaller:
> > I suppose if the tests are executed in alphabetical order the 000 causes those tests to execute first (?). Is that desirable?
> 
> One of the points of redoing the tests was that they can all now be run in parallel, so I guess the order is (should be) irrelevant.

I see, groovy.  BTW, when running under GitHub Actions I think we typically only have two cores available.  This may vary, however; I see that some people have put stuff in their YAML file to determine how many cores there are (e.g., https://github.com/SimenB/github-actions-cpu-cores).  I don't know how you decide how many cores to spawn jobs for, but maybe this is useful info.  :->

PR #147, opened 2021-01-23
Update to tskit C 0.99.10

hyanwong:
Currently not compiling because of `tsk_vargen_t->samples` missing in new tskit version. This is actively being discussed on Slack.

### DIFF
--- a/treerec/tskit/convert.c
+++ b/treerec/tskit/convert.c
@@ -25,7 +25,6 @@
 
 #include <stdio.h>
 #include <string.h>
-#include <assert.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <math.h>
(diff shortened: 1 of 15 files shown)


hyanwong:
OK, this is updated to tskit 0.99.10 @bhaller, and so now compiles. Do you want to (a) make any extra changes re whether to use `ts->samples` or whatever (I think the conclusion was to leave as-is though?), and (b) run your full test suite on it?

bhaller:
I am going to do an optimistic merge; I don't want to stall progress doing all the testing first, which I'm sure will pass, and if it *doesn't* pass, well, we will cross that bridge when we come to it.  :->  So, your (a) will be dealt with separately by me (there is a change to be made but it is orthogonal to this), and I will kick off your (b) as soon as I've merged, and leave the tests running today while I deal with some unrelated stuff.  :->  Thanks for the PR!

bhaller:
FYI, the full test suite with crosscheck passed, all is well.  I just committed the tweak to the tsk_vargen_t usage, which was part of what was tested, so we can all forget about that issue now too.  Full speed ahead.  :->

PR #149, opened 2021-01-27
Switch to tsk_flags_t for flags passed in to tskit C funcs

hyanwong:
Also clarify use of "retained" in comments about individuals.


bhaller:
Good fixes, thanks!

PR #150, opened 2021-01-27
Fix retain bug

hyanwong:
I forgot to add to the list of remembered_genomes_ if the individual was present, but not previously remembered.

Thanks for spotting there was a problem, @petrelharp !


bhaller:
Aha, good edge case spotting.

PR #155, opened 2021-02-03
Activate "keep unary in individuals"

hyanwong:
As it might be a little while until the tskit fix for keep unary in individuals makes it into a release, I have hacked in the fix by hand in the PR. I have also renamed some of the tests as per https://github.com/MesserLab/SLiM/pull/142#issuecomment-768457197 to replace `000` with `___` which I think is less confusing. There's also a new suite of python tests intended to test the results of specific SLiM recipes, rather than more general tests of any TS produced by SLiM.

I'm now move on to documenting this in pyslim, so that @petrelharp can release a compatible version.

### DIFF
--- a/treerec/tests/conftest.py
+++ b/treerec/tests/conftest.py
@@ -135,11 +135,12 @@ class OutputResult:
             fields = line.split()
             assert fields[0].startswith("#Individual:")
             store = fields[0][len("#Individual:"):]
-            assert store in ("remember", "retain", "output")
+            assert store in ("remember", "retain", "retain_if_unary", "output")
             pedigree_id = int(fields[1])
-            if pedigree_id in slim and store == "retain":
-                # We have a duplicate: "remember"ed takes priority
-                continue
+            if pedigree_id in slim:
+                # We have a duplicate:
+                if slim[pedigree_id].type == "remember":
+                    continue  # "remember"ed takes priority: do not overwrite
             slim[pedigree_id] = SLiMindividual(
                 type=store,
                 population=int(fields[2]),
@@ -166,7 +167,7 @@ def run_slim(recipe, run_dir, recipe_dir="test_recipes"):
     these directories, which contain the files on which tests should be run
     """
     script_dir = os.path.dirname(os.path.realpath(__file__))  # Path to this file
-    full_recipe = os.path.abspath(os.path.join(recipe_dir, recipe))
+    full_recipe = os.path.abspath(os.path.join(script_dir, recipe_dir, recipe))
     assert os.path.isdir(run_dir)  # should have been created by caller
     cmd = ["slim", "-s", "22", "-d", f"RUN_DIR=\"{run_dir}\"", full_recipe]
     print(f"Running {cmd} in dir '{run_dir}', errors etc to 'SLiM_run_output.log'")
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -5043,7 +5043,9 @@ void SLiMSim::SimplifyTreeSequence(void)
 	if (ret < 0) handle_error("tsk_table_collection_deduplicate_sites", ret);
 	
 	// simplify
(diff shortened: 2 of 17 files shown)


bhaller:
This all sounds great, thanks!  Is it ready to merge?  Should @petrelharp review it, or is there no need?

hyanwong:
> This all sounds great, thanks! Is it ready to merge? Should @petrelharp review it, or is there no need?

I think it might be helpful, if he has time. Perhaps at the same time as looking at the pyslim changes I'm about to push to keep this in sync?

treerec/tests/conftest.py
```diff
@@ -135,11 +135,12 @@ def individual_output(self):
...
-            assert store in ("remember", "retain", "output")
+            assert store in ("remember", "retain", "retain_if_unary", "output")
             pedigree_id = int(fields[1])
-            if pedigree_id in slim and store == "retain":
-                # We have a duplicate: "remember"ed takes priority
-                continue
+            if pedigree_id in slim:
+                # We have a duplicate:
+                if slim[pedigree_id].type == "remember":
+                    continue  # "remember"ed takes priority: do not overwrite
```
petrelharp, line 143: not getting the whole picture here, but do we also need a "retain takes priority over retain if unary" here?

petrelharp: oh, I see, we only check below if this starts with retain, so it doesn't matter? still, it'd be nice to get the right thing in here.

hyanwong: We never have both `retain_if_unary` and `retain` in the same simulation, so it's never an issue.

petrelharp: ah, well, perhaps we should have them both in the same simulation?

hyanwong: Since we only specify "retainCoalescentOnly" for the entire TS recording, we have no way of doing this. But I can change the logic in case we ever do implement something similar? I can't see how it would work though!

petrelharp: Sorry, maybe I'm just confused, but can't we do `retainCoalescentOnly=T` up top, and then insert some calls to `treeSeqRememberIndividuals(..., permanent=F)` in the recipe?

hyanwong: Yes we can do that. If so, we need to mark all retained individuals as "retain". If, at the top, we use `retainCoalescentOnly=F`, we save details of all retained individuals under `retain_if_unary`.

hyanwong: Note that "retain" and "retain_if_unary" are only convenience flags in the text-save of the individuals, for testing. And also note that any individuals we zap with `treeSeqRememberIndividuals(..., permanent=T)` are marked as "remember" which always overrides either "retain" marking.

petrelharp: I think I'm being slow here, but: we can do `retainCoalescentOnly=T`, and then only mark SOME of the individuals to retain; this will produce a mix of `retain` and `retain_if_unary`. Ah, I see - if `retainCoalescentOnly=T`, then should `retain` have no effect? (Still, we could be checking nothing weird happens.)

hyanwong: I think there's some confusion here. `retainCoalescentOnly=T` only applies during `treeSeqInitialize()`. So either it's on for the whole simulation, or off for the whole simulation.

If "on", then any node permanently remembered is output with a "remember" tag, and and node temporarily retained is output with a "retain" flag.

If "off", then any node permanently remembered is output with a "remember" tag, and and node temporarily retained is output with a "retain_if_unary" flag.

petrelharp: ack, sorry, I think I need some more coffee now.

bhaller:
I see there was some review by @petrelharp.  Is that complete?  Is this ready to merge?  Don't mean to nag, just wondering if this has stalled for no reason.  :->

hyanwong:
AFAIK this just needs @petrelharp 's sign-off. it's linked to https://github.com/tskit-dev/pyslim/pull/125 which needs, I think, one extra test, is that right @petrelharp ?

Issue #139, opened 2021-01-11
Allow `keep_unary` to be passed to simplification funcs


hyanwong:
I can imagine that it would be very handy to be able to output in  tree sequence all the intermediate nodes produced by SLiM - i.e. not just the coalescence points, but all the "unary nodes" too. This would be somewhat intermediate between "saving everybody" and the normal output. Individuals would still not be saved, of course, unless they happen to be ["retained"](#40) - this is just about saving the node info (genomes) themselves.

The way I imagine this working would be to add an optional `keep_unary` arg to `treeSeqSimplify()`, which could be T, F, or NULL (default). There would also be a `keep_unary` option to `initializeTreeSeq()` that would default to `F` and would set the default `keep_unary` flag during automatic simplification. This value would be the default used when calling `treeSeqSimplify()` with no params.

What do you think @petrelharp ? I guess you're the one doing "remember everyone" simulations.

bhaller:
As I mentioned in the review on #25, let's try to envision the ideal final design for this stuff.  Is it a `keepUnary=T` option on `initializeTreeSeq()` and a `permanent=F` option on `treeSeqRememberIndividuals()`?  In other words, are these two features completely orthogonal?  Or is there some interaction between them, and some better API design?  Because `keepUnary=T` strikes me as a kind of "remembering" or "retaining" too, but we're proposing handling it in a very different way.  It could be handled instead as another type of flag passed to `treeSeqRememberIndividuals()`, right?  A flag saying "mark these individuals as being kept even if their nodes are unary"?  I don't see a clear problem with what has been proposed, but I want to make sure we're going to arrive at a final design that we won't regret later.

hyanwong:
I see them as completely orthogonal, which is why I opened up a new issue. I guess you could think of this as a sort of "remembering", but of *nodes* not individuals. The distinction is very clear in tree-sequence world, and I think that the "keep_unary" function should automatically play nicely (and logically) with the `permanent=F` option in #25 . But let's see what @petrelharp thinks too?

petrelharp:
Not even so much orthogonal, but copmplementary: if you do `keep_unary=T` and you Retain all the individuals, then you'll get the individuals corresponding to retained unary nodes as well. (However, I'd imagine that would be nearly the same as just Remembering everyone.)

hyanwong:
> (However, I'd imagine that would be nearly the same as just Remembering everyone.)

[...]

Issue #151, opened 2021-01-27
add treerec tests to things run automatically by github


petrelharp:
This just means adding something to `.github/workflows/tests.yml` I think.

bhaller:
This would go in the same place where it does `./slim -testSLiM`.  Note that there are two such places, one for Debug and one for Release.  I'd do this myself but I don't know how to run the python tests!  :->

PR #158, opened 2021-02-20
Keep unary in individuals activate

petrelharp:
Well, this is awkward. I made some minor tweaks to #155:
- fixed a spaces-to-tabs error,
-  renamed the label from testing from "keep_if_unary" to "keep_even_if_unary" (since the previous name was confusing me)
- fixed up a thing to do with priority in the testing suite that didn't affect anything but is more correct now

And, rebased. But since I rebased I can't just make a PR to #155. So, this PR includes #155, and my recommendation is to merge this one and close-without-merging that one.


bhaller:
OK, done, thanks.  So, where does this leave us?  Did this PR close #139, or is there more to be done?  #151 would also be good to do; I think that should be very simple to fix.

petrelharp:
Yep, #139 can be closed! I've got a PR for #151, we'll see if it works...

PR #159, opened 2021-02-20
github run pytests: closes #151

### LINKED ISSUE #151: add treerec tests to things run automatically by github

petrelharp:
Let's see if this works!

### DIFF
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -6269,14 +6269,6 @@ void SLiMSim::AddIndividualsToTable(Individual * const *p_individual, size_t p_n
 					   == (p_tables->individuals.metadata_offset[tsk_individual + 1]
 						   - p_tables->individuals.metadata_offset[tsk_individual])));
 			
-			// BCH 4/29/2019: This assert is, we think, not technically necessary – the code
-			// would work even if it were violated.  But it's a nice invariant to guarantee,
-			// and right now it is always true.
-			assert((p_tables->nodes.individual[ind->genome1_->tsk_node_id_]
-					 == tsk_individual)
-				   && (p_tables->nodes.individual[ind->genome2_->tsk_node_id_]
-					   == tsk_individual));
-			
 			// It could have been previously inserted but not with the 
 			// SLIM_TSK_INDIVIDUAL_REMEMBERED flag: if so, it now needs adding to the
 			// list of remembered_genomes_
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,6 +23,7 @@ jobs:
         # see https://github.com/actions/virtual-environments/issues/2486 for more info
         # also relevant: https://github.com/actions/virtual-environments/issues/2381
         os: [ubuntu-18.04, ubuntu-20.04, macos-10.15]
+        python: [ 3.8 ]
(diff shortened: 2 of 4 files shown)


bhaller:
Looks like it failed: "Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'".

petrelharp:
heh, whoops

petrelharp:
I'm asking this to run with development pyslim, FYI. I think that's the right thing to do here.

bhaller:
So, the tests.yml builds both a Debug and a Release build of SLiM.  It's not clear to me which one your tests will run against.  How is it finding the `slim` binary to execute?

core/slim_sim.cpp
```diff
@@ -6269,14 +6269,6 @@ void SLiMSim::AddIndividualsToTable(Individual * const *p_individual, size_t p_n
...
 						   - p_tables->individuals.metadata_offset[tsk_individual])));
 			
-			// BCH 4/29/2019: This assert is, we think, not technically necessary – the code
-			// would work even if it were violated.  But it's a nice invariant to guarantee,
-			// and right now it is always true.
-			assert((p_tables->nodes.individual[ind->genome1_->tsk_node_id_]
-					 == tsk_individual)
-				   && (p_tables->nodes.individual[ind->genome2_->tsk_node_id_]
-					   == tsk_individual));
-
```
petrelharp, line 6279: This bit of code is duplicated above, and there doesn't seem to be the possibility that the node tables are changed between there and here?

bhaller: If you're referring to the line above where it does `p_tables->nodes.individual[ind->genome1_->tsk_node_id_] = tsk_individual;`, that is in a different branch of the if/else statement, isn't it?

petrelharp:
Huh, this still fails on macos by not finding gsl when installing msprime, but the exact same coide, AFAICT, works for installing msprime for tskit?!? Perhaps @benjeffery knows what's going on?

core/slim_sim.cpp
petrelharp: Sorry, I meant ["below", here](https://github.com/MesserLab/SLiM/pull/159/files/d2b23c05f7e0e2a1642154a1f22d0560e1dbb113#diff-0ae7a04e87e1ce83b282da28825cc9492ad328d2d4a6d0343fb24f61137de991R6294).

petrelharp: Figured I'd keep the consistency check that's at the end.

petrelharp:
IT WORKS!!!!!!! :tada: I think this is ready to merge!

If you're wondering what all this conda cache stuff is doing, so am I, but I think the answer is that "it's making things work" and we can thank Ben Jeffery for figuring it out.

core/slim_sim.cpp
bhaller: Ah, I see.  Yes, the first check seems redundant, I suppose.  I can't remember why we did that.

bhaller:
Wow, that is impressively gross.  :->  I guess understanding it isn't really necessary – it's just CI glue.  Congrats on slogging through to the finish line!

PR #160, opened 2021-02-20
actually use the cache

petrelharp:
I guess certain parts of the cache are not strictly necessary...


petrelharp:
It seems to have run and passed the tests, so you can merge this, too (sorry I missed this before).

PR #164, opened 2021-03-07
Freshen the readme

bryce-carson:
I think this version of the readme is more readable and usable. The previous version was a little _uneasy_ on the eyes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,17 @@
+<p align="center">
+	<img alt="Screenshot of SLiMgui running on OS X." height="45%" width="45%" src="https://messerlab.files.wordpress.com/2016/02/slimgui_screenshot.jpg?w=600"/>
+</p>
+
 **SLiM on GitHub:** ![SLiM on GitHub Actions:](https://github.com/MesserLab/SLiM/workflows/tests/badge.svg)
 
 **SLiM on Travis-CI:** ![SLiM on Travis-CI:](https://travis-ci.com/MesserLab/SLiM.svg?branch=master)
 
 **SLiM on COPR:** [![Copr build status](https://copr.fedorainfracloud.org/coprs/bacarson/SLiM-Selection_on_Linked_Mutations/package/SLiM/status_image/last_build.png)](https://copr.fedorainfracloud.org/coprs/bacarson/SLiM-Selection_on_Linked_Mutations/package/SLiM/)
 
----------------------------------------------------------------------------------------------
-PLEASE NOTE THAT THE GITHUB VERSION OF SLiM IS PRERELEASE SOFTWARE THAT IS UNDER ACTIVE DEVELOPMENT!
----------------------------------------------------------------------------------------------
-It is strongly recommended that all end users of SLiM use the current release version, available at [http://messerlab.org/slim/](http://messerlab.org/slim/).
----------------------------------------------------------------------------------------------
- 
+This GitHub repository hosts the <em>upstream, development head version</em> of SLiM and SLiMgui and should <strong>not</strong> be used by end users.
+
+The <strong><em>release</em></strong> version of SLiM and SLiMgui are available at [http://messerlab.org/slim/](http://messerlab.org/slim/).
+
 
 License
 ----------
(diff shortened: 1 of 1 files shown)


bhaller:
Hi @bryce-carson.  Thanks!  Is there a way for me to see how Git Hub will render this prior to accepting the PR?

bryce-carson:
Yes, if you look at the rendered readme on my fork you'll see how it is displayed on Web.

The GitHub app differs only in not having the folded build instructions at the bottom.

bhaller:
Got it, thanks.  Copyright date should be fixed to end in 2021.  Overall, an improvement for sure.  I think I'd vote for removing the icons to the left of the section headers; let's go for a little more formality in the appearance.  I'd move the "binary packages" section to be above "compilation from sources", since that is the more common/recommended case.  And I'd actually just get rid of the "compiling from sources" folded instructions completely; just referring to the manual for that is better, that's where the definitive instructions live now.  That can refer explicitly to chapter 2 of the manual.  The SLiMgui snapshot should be of the new QtSLiM, but I see you're pulling that from the Messer Lab page, so I need to fix that.  :->  BTW the Fedora Copr status shows as a "?" icon for me, not sure what's up with that.  Sound good to you?

bryce-carson:
Yeah, that sounds fine. Feel free to merge with changes, or I will get around to it. :)

bryce-carson:
I'm not sure what was up with Copr, I think it might have been "down" last night around the time you were looking at it; I was unable to access Copr about the same time.

bhaller:
OK, I looked it over again.  A couple of little nits that I can change after merging, if you're feeling done with this PR :->.

- For "End-users should **not** use these sources" I would substitute: "**End users should generally _not_ use these sources; they may contain serious bugs, or may not even compile.**"  I want to really emphasize that, because lots of biologists have no idea what GitHub is, how releases work, etc., and may use the GitHub head version without understanding what they're getting into.

- For "The release version of SLiM and SLiMgui are available" change to "is available"; one version, one product, one release, thus singular.  :->

- You have "Windows 10 (WSL & GSL)".  I'm used to "GSL" being the "GNU Scientific Library".  What does it stand for in this context?  Would it be less confusing to just say "WSL", since that seems to be the term that people use for it?

Again, little nits; I'll do this myself if you prefer, in which case I can accept this PR now.  Thanks!

bryce-carson:
Those first two changes are welcomed.

For the third point, I included that to refer to my "easier" way to work with WSL, using GWSL (Graphical WSL, it's on the Windows Store). Since it's not actually an installer we ship, and it just uses the debian script with some extra steps we can remove it and rely on the manual instead to show that using SLiM on Windows 10 is possible.

bryce-carson:
openSUSE is a Linux Distribution. I was referring to the conditional Requires: and BuildRequires: statements for the packages needed to use SLiM on that Distribution.

bhaller:
> Those first two changes are welcomed.
> 
> For the third point, I included that to refer to my "easier" way to work with WSL, using GWSL (Graphical WSL, it's on the Windows Store). Since it's not actually an installer we ship, and it just uses the debian script with some extra steps we can remove it and rely on the manual instead to show that using SLiM on Windows 10 is possible.

Ah – your text says "GSL" but it sounds like it ought to say "GWSL"?

bhaller:
> openSUSE is a Linux Distribution. I was referring to the conditional Requires: and BuildRequires: statements for the packages needed to use SLiM on that Distribution.

Yep, I got that.  :->  I just meant that I've never heard of openSUSE before, don't know anything about it beyond that it is a Linux distro, etc.  But the more the merrier!  :->

bhaller:
> Those first two changes are welcomed.
> 
> For the third point, I included that to refer to my "easier" way to work with WSL, using GWSL (Graphical WSL, it's on the Windows Store). Since it's not actually an installer we ship, and it just uses the debian script with some extra steps we can remove it and rely on the manual instead to show that using SLiM on Windows 10 is possible.

Am I supposed to accept this PR and make my proposed changes, or am I waiting for a new commit?  And is "GWSL" correct, or is "GSL" short for that?  :->

bryce-carson:
> > Those first two changes are welcomed.
> > For the third point, I included that to refer to my "easier" way to work with WSL, using GWSL (Graphical WSL, it's on the Windows Store). Since it's not actually an installer we ship, and it just uses the debian script with some extra steps we can remove it and rely on the manual instead to show that using SLiM on Windows 10 is possible.
> 
> Am I supposed to accept this PR and make my proposed changes, or am I waiting for a new commit? And is "GWSL" correct, or is "GSL" short for that? :->

Ah, yes. You can accept the PR and make the changes, please.

GWSL is the correct name. If we want, we can keep that cell in the table to refer to the installation instructions md file we pushed today. :)

bhaller:
Pushed a few cleanup edits just now, including linking to your GWSL instructions from the installation table.  This seems pretty much done.  Thanks again!

Issue #169, opened 2021-03-18
Add `name` and `description` to population metadata


petrelharp:
msprime v1.0 by default tries to use "name" and "description" properties pulled from population metadata. This is very convenient, as it lets one specify demography using population names instead of IDs. And, it's a good idea. I think we should add these to the population metadata (and am happy to do it sometime before the next release).

bhaller:
Sure.  I guess the "name" for p1 would be "p1", etc.?  Are you proposing to add a "description" to SLiM subpops?  That is not unreasonable, but would require new Eidos API.

petrelharp:
Right - I'm just proposing using `p1`, etcetera, and not using the "description" field.

And: we'd avoid some future headaches if we switched population metadata to using JSON instead of a struct, and set additionalProperties=True in the schema. (I'm not suggesting this for other metadata, for performance, though.)

bhaller:
> Right - I'm just proposing using `p1`, etcetera, and not using the "description" field.

OK, that seems like a good idea.

> And: we'd avoid some future headaches if we switched population metadata to using JSON instead of a struct, and set additionalProperties=True in the schema. (I'm not suggesting this for other metadata, for performance, though.)

Interesting.  I guess if we increment the overall file version at the same time, then we know whether we expect the metadata to be JSON or struct, so we can safely read old files.  Sounds like a bit of a pain, but do-able.  Sure.  I think we have kind of an existing example of how to handle this, with MutationMetadataRec versus MutationMetadataRec_PRENUC, although it's not quite the same since JSON is not involved.  I guess for this proposed change we would change the name of SubpopulationMetadataRec to SubpopulationMetadataRec_PREJSON or some such, and use that for the backward-compatibility code, and we wouldn't have a SubpopulationMetadataRec at all; we'd have JSON encoding/decoding stuff instead, to/from the bytes in the metadata field.  Should be straightforward enough, let me know if you get stuck?

grahamgower:
Maybe it would be reasonable to have `name` and `description` fields in the `Subpopulation` class, and write those out to the tree sequence metadata? As you say, the name could just default to "p1", etc. if not set by the user.

grahamgower:
Oh, and if you end up pulling in a json parser, I might have future JSON-related feature requests. :)

bhaller:
[...]

PR #171, opened 2021-03-18
remove subpop number check

### LINKED ISSUE #168: remove check on number of input populations?
I've run into the following error: on [this line](https://github.com/MesserLab/SLiM/blob/91beaaa5fd728cbf28fd2f54a8d3360dc5af9626/core/slim_sim.cpp#L7939) when SLiM is reading in a trees file for a WF model, it checks that the number of populations-with-SLiM-metadata in the trees file matches the number of subpops, that I'm guessing is created by iterating over the individuals and creating the populations. However: suppose that I (1) run a SLiM simulation with 4 populations; (2) write out the trees file and do something to it, including removing all the individuals except in one population; (3) want to read it into SLiM and simulate a WF model with just one population.  This will (I think? haven't double-checked) raise that error.

I could, in addition to removing individuals, also remove the corresponding entries in the population table. But those populations are still valid - ancestral nodes refer to them - so I guess I'd not remove their entries, just remove their metadata.

But, I don't think we need this check at all? I think it was just a consistency check, but really, the "population table" should contain information about not just the current generation, but also ancestral ones.

And: where I've actually encountered this problem is in recapitation. It's natural to want to add new 'ancestral' populations in recapitation, and it's being difficult to set their metadata appropriately. We can deal with that somehow, but I'm hoping we can do away with this check for the reasons above.

petrelharp:
I *think* this correctly just doesn't add a new subpopulation for WF models with no individuals. However, I am not sure about the bigger ramifications, and also probably the `continue;` is not the most elegant way to do it.

Will close #168.

### DIFF
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -7923,23 +7923,6 @@ void SLiMSim::__ConfigureSubpopulationsFromTables(EidosInterpreter *p_interprete
 	tsk_population_table_t &pop_table = tables_.populations;
 	tsk_size_t pop_count = pop_table.num_rows;
 	
-	// do a quick sanity check that the number of non-empty rows equals the expected subpopulation count, in WF models
-	if (model_type_ == SLiMModelType::kModelTypeWF)
-	{
-		tsk_size_t nonempty_count = 0;
-		
-		for (tsk_size_t pop_index = 0; pop_index < pop_count; pop_index++)
-		{
-			size_t metadata_length = pop_table.metadata_offset[pop_index + 1] - pop_table.metadata_offset[pop_index];
-			
-			if (metadata_length > 0)
-				++nonempty_count;
-		}
-		
-		if ((size_t)nonempty_count != population_.subpops_.size())
-			EIDOS_TERMINATION << "ERROR (SLiMSim::__ConfigureSubpopulationsFromTables): subpopulation count mismatch; this file cannot be read." << EidosTerminate();
-	}
-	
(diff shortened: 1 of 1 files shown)


petrelharp:
To get a MWE, you can run:
```
initialize() {
	initializeTreeSeq();
	initializeMutationRate(1e-7);
	initializeMutationType("m1", 0.5, "f", 0.0);
	initializeGenomicElement(g1, 0, 10);
	initializeRecombinationRate(1e-8);
}
1 {sim.addSubpop("p1", 2); }
1 late() { sim.treeSeqOutput("slim.trees"); }
```
and then run
```
import tskit, pyslim
tables = tskit.load("slim.trees").tables
md = pyslim.default_slim_metadata('population')
tables.populations.add_row(md)
tables.dump("slim_plus.trees")
```
and then back in SLiM
```
initialize() {
	initializeTreeSeq();
	initializeMutationRate(1e-7);
	initializeMutationType("m1", 0.5, "f", 0.0);
	initializeGenomicElement(g1, 0, 10);
	initializeRecombinationRate(1e-8);
}

1 {
   sim.readFromPopulationFile("slim_plus.trees");
}

1 late() { sim.treeSeqOutput("slim.trees"); }
```

bhaller:
I think the strategy of this PR is fine, and I don't *think* it will have any negative downstream effects.  I don't know whether we preserve the unreferenced subpop on write, and the problem hypothesized in #169 probably does exist.  But those are separate issues.  I fixed this with a commit from my side, to add a little more commenting, and to avoid tab/space issues with the diffs here, but the fix is the same.

PR #172, opened 2021-03-18
write out unused population table entries unchanged

petrelharp:
Here's a draft of leaving alone all population table entries that we don't use, instead of zeroing them out, like we did before.

### DIFF
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -6395,16 +6395,19 @@ void SLiMSim::FixAliveIndividuals(tsk_table_collection_t *p_tables)
 
 void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
 {
-	// This overwrites whatever might be previously in the population table
-	tsk_population_table_clear(&p_tables->populations);
-	
-	// we will write out empty entries for all unused slots, up to largest_subpop_id_
-	// this is because tskit doesn't like unused slots; slot indices must correspond to subpop ids
-	slim_objectid_t last_subpop_id = -1; // population_.largest_subpop_id_;
-	
+	int ret;
+	tsk_id_t tsk_population_id;
+	tsk_population_table_t *population_table_copy;
+	population_table_copy = (tsk_population_table_t *)malloc(sizeof(tsk_population_table_t));
+	ret = tsk_population_table_copy(&p_tables->populations, population_table_copy, 0);
+	if (ret != 0) handle_error("WritePopulationTable tsk_population_table_copy()", ret);
+	ret = tsk_population_table_clear(&p_tables->populations);
+	if (ret != 0) handle_error("WritePopulationTable tsk_population_table_clear()", ret);
+
+	slim_objectid_t last_subpop_id = population_table_copy->num_rows - 1;
 	for (size_t j = 0; j < p_tables->nodes.num_rows; j++)
 		last_subpop_id = std::max(last_subpop_id, p_tables->nodes.population[j]);
-	
+
 	// write out an entry for each subpop
 	slim_objectid_t last_id_written = -1;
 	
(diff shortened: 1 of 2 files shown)


core/slim_sim.cpp
```diff
@@ -6381,16 +6381,10 @@ void SLiMSim::FixAliveIndividuals(tsk_table_collection_t *p_tables)
 
 void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
 {
-	// This overwrites whatever might be previously in the population table
+    tsk_population_table_t population_table_copy;
```
bhaller, line 6384: tabs instead of spaces would make these diffs easier to read...

core/slim_sim.cpp
```diff
@@ -6402,8 +6396,14 @@ void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
 		// first, write out empty entries for unused subpop ids before this one
 		while (last_id_written < subpop_id - 1)
 		{
-			tsk_id_t tsk_population = tsk_population_table_add_row(&p_tables->populations, (char *)&last_subpop_id, (uint32_t)0);	// the address is unused
-			if (tsk_population < 0) handle_error("tsk_population_table_add_row", tsk_population);
+            tsk_population_t tsk_population_object;
+            tsk_population_table_get_row(&population_table_copy, last_id_written, &tsk_population_object);
```
bhaller, line 6400: don't you need to check whether a row even exists at this index?

core/slim_sim.cpp
```diff
@@ -6435,20 +6435,23 @@ void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
...
 	
 	// finally, write out empty entries for the rest of the table; empty entries are needed
 	// up to largest_subpop_id_ because there could be ancestral nodes that reference them
-	while (last_id_written < last_subpop_id)
+	while (last_id_written < (slim_objectid_t) population_table_copy.num_rows)
 	{
-		tsk_id_t tsk_population = tsk_population_table_add_row(&p_tables->populations, (char *)&last_subpop_id, (uint32_t)0);	// the address is unused
-		if (tsk_population < 0) handle_error("tsk_population_table_add_row", tsk_population);
+        tsk_population_t tsk_population_object;
+        tsk_population_table_get_row(&population_table_copy, last_id_written, &tsk_population_object);
```
bhaller, line 6451: ditto

bhaller (commented):
Here are some comments.  Please add an entry to the VERSIONS file for this change.  Thanks!

petrelharp:
Ok, I think this is good! With minimal testing so far, though.

bhaller:
> Ok, I think this is good! With minimal testing so far, though.

Looks like the compile is failing?  "‘tsk_population_id’ was not declared in this scope"

petrelharp:
Hm - it's got a point, but my compiler didn't complain? Well, fixed now, hopefully?

petrelharp:
Er, no, whoops...

petrelharp:
Ok, looks like I got it for reals this time.

core/slim_sim.cpp
```diff
@@ -6449,22 +6466,35 @@ void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
...
+			if (tsk_population_id < 0) handle_error("tsk_population_table_add_row", tsk_population_id);
+		} else {
+			tsk_population_id = tsk_population_table_add_row(
+					&p_tables->populations,
+					NULL, 0);
+			if (tsk_population_id < 0) handle_error("tsk_population_table_add_row", tsk_population_id);
+		}
+		assert(tsk_population_id == last_id_written);
 	}
+	free(population_table_copy);
```
bhaller, line 6497: isn't there a tskit free-the-internals-of-this function we're supposed to call before free() here?

core/slim_sim.cpp
```diff
@@ -6416,9 +6419,23 @@ void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
...
+				if (ret != 0) handle_error("WritePopulationTable tsk_population_table_get_row()", ret);
+				tsk_population_id = tsk_population_table_add_row(
+						&p_tables->populations,
+						tsk_population_object.metadata,
+						tsk_population_object.metadata_length);
+				if (tsk_population_id < 0) handle_error("tsk_population_table_add_row", tsk_population_id);
+			} else {
+				tsk_population_id = tsk_population_table_add_row(
+						&p_tables->populations,
+						NULL, 0);
```
bhaller, line 6435: just to check: the old code passed `(char *)&last_subpop_id` instead of NULL when adding new empty rows.  is the switch to NULL definitely OK?

bhaller:
Two final questions from code review above; once those are resolved (perhaps through inaction :->) I'll merge, let me know.

core/slim_sim.cpp
petrelharp: NULL is what we usually use in the tskit code so unless C++ is for some reason different, then yes?

petrelharp: See for instance here: https://github.com/tskit-dev/tskit/blob/f1fd80462ae1b0e211d651440b88c224b7cda393/c/tests/test_tables.c#L168

bhaller:
OK, merged, thanks!

petrelharp:
Good catch on the `free`!

PR #197, opened 2021-06-11
Check integrity

### LINKED ISSUE #193: enable SLiMSim::CheckTreeSeqIntegrity()
SLiMSim::CheckTreeSeqIntegrity() has this code:

	// Here we call tskit to check the integrity of the tree-sequence tables themselves – not against
	// SLiM's parallel data structures (done in CrosscheckTreeSeqIntegrity()), just on their own.
	// At present this is commented out because tsk_table_collection_check_integrity() does checks
	// that we always fail.  In particular, we have no population table in our table collection while
	// we are running, so we always get a TSK_ERR_POPULATION_OUT_OF_BOUNDS error.  When tskit is
	// made more flexible, in terms of being able to turn off checks that we do not want, we will
	// enable this code with the appropriate flags.  See https://github.com/tskit-dev/tskit/issues/593
	
	/*int ret = tsk_table_collection_check_integrity(&tables_, 0);
	if (ret < 0) handle_error("tsk_table_collection_check_integrity()", ret);*/

I think tskit now allows us to customize our checks?  @petrelharp

petrelharp:
This closes #193 - and seems to work?

This PR has the commit from #196 as well, so that one should be merged before this one (then this show as a clean diff with just the one-line change).

### DIFF
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -7627,8 +7627,8 @@ void SLiMSim::CheckTreeSeqIntegrity(void)
 	// made more flexible, in terms of being able to turn off checks that we do not want, we will
 	// enable this code with the appropriate flags.  See https://github.com/tskit-dev/tskit/issues/593
 	
-	/*int ret = tsk_table_collection_check_integrity(&tables_, 0);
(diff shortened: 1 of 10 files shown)


bhaller:
OK, looks good; I'll merge this once #196 is merged.

PR #204, opened 2021-07-22
Update GitHub Action to use latest GCC (v11).

grahamgower:
SLiM v3.6 has a build issue with GCC-11 (fixed by https://github.com/MesserLab/SLiM/commit/7964f7911e1665a1ba0783d9136b3d3bff06d931). See https://www.gnu.org/software/gcc/gcc-11/porting_to.html for details of header changes. By testing on GCC-11, we help ensure there are no regressions in the future.

Thanks to Clemens Schmid (@nevrome) for alerting me to this via the Arch Linux package.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
             gcc: 4.8
           # Ubuntu with the most recent GCC.
           - os: ubuntu-20.04
-            gcc: 10
+            gcc: 11
 
     steps:
       - name: Checkout
@@ -150,7 +150,7 @@ jobs:
           # Ubuntu with the most recent Qt and GCC.
           - os: ubuntu-20.04
             qt: 5.15.2
-            gcc: 10
+            gcc: 11
           # MacOS with oldest supported Qt.
           - os: macos-10.15
             qt: 5.9.5


bhaller:
I wonder if it might be a good idea to test both GCC 10 and GCC 11?  I don't have a strong opinion on that; but it would give us more comprehensive coverage, at no obvious cost.  Maybe our coverage with Github Actions ought to be a bit broader in general; our travis-ci testing is completely kaput now, unfortunately, because travis itself went down the tubes after being bought, as far as I can tell, at least for open-source projects like ours.  So for right now GA is our only CI.  Anyhow, I'll accept this PR as-is if you think it's best, but just wanted to raise the question.

grahamgower:
Well that's really up to you and what you think is important. The CI is running at ~18 minutes right now, which is already a fairly long time to wait for feedback. So probably you'll want to make a trade off for the CI time against testing other configurations, like more OS versions (e.g. macos 11 when that becomes viable) or some other dependency that causes problems (maybe cmake?).  My opinion is that there's diminishing returns on testing lots of different versions of GCC --- the oldest and most recent versions are going to expose the majority of problems, and it's pretty unlikely to see a build issue with an intermediate version that isn't also an issue with the older or newer versions.

Perhaps more importantly, is there anything that the travis CI was testing that isn't being tested by the github action?

bhaller:
> Well that's really up to you and what you think is important. The CI is running at ~18 minutes right now, which is already a fairly long time to wait for feedback. So probably you'll want to make a trade off for the CI time against testing other configurations, like more OS versions (e.g. macos 11 when that becomes viable) or some other dependency that causes problems (maybe cmake?).

Well, it seems like all the CI tasks run concurrently, not consecutively, so I don't think it would increase the average time to completion for the CI (unless GCC 10 is substantially slower than other GCC versions).  Am I misunderstanding?

> My opinion is that there's diminishing returns on testing lots of different versions of GCC --- the oldest and most recent versions are going to expose the majority of problems, and it's pretty unlikely to see a build issue with an intermediate version that isn't also an issue with the older or newer versions.

OK, fair enough.  We don't want to get into trying to comprehensively test every version, so oldest and newest makes sense, I guess.  OK.

> Perhaps more importantly, is there anything that the travis CI was testing that isn't being tested by the github action?

I mean, slightly different variations and configurations than we have now in GA, but nothing that seems particularly essential.

OK, I think this is fine, just wanted to hear your thoughts.  I'll merge it now.  Thanks!

grahamgower:
> > Well that's really up to you and what you think is important. The CI is running at ~18 minutes right now, which is already a fairly long time to wait for feedback. So probably you'll want to make a trade off for the CI time against testing other configurations, like more OS versions (e.g. macos 11 when that becomes viable) or some other dependency that causes problems (maybe cmake?).
> 
> Well, it seems like all the CI tasks run concurrently, not consecutively, so I don't think it would increase the average time to completion for the CI (unless GCC 10 is substantially slower than other GCC versions). Am I misunderstanding?

Ah, you're absolutely right! There is a job queue, but it's somewhere hidden inside google's github farm. So it's not obvious how extra test configurations would affect the test turnaround. \*shrug\*

PR #210, opened 2021-09-17
sort/deduplicate on the table copy; fix #209

### LINKED ISSUE #209: WriteTreeSeq modifies tables
Over in #208, @bhaller  pointed out that

> (Incidentally – WriteTreeSequence() [calls tsk_table_collection_sort() and tsk_table_collection_deduplicate_sites()]( https://github.com/MesserLab/SLiM/blob/master/core/slim_sim.cpp#L7294) on the simulation's tables, before it makes a copy of the simulation's tables with tsk_table_collection_copy(), when simplify=F, so calling treeSeqOutput(simplify=F) in Eidos has side effects on the simulation's tables; do we really want to be doing that before we copy the tables, or is that a separate bug?)

The deduplicate sites should be harmless; but the `sort` is a bit worrisome. On the one hand, we do both of these [during simplification](https://github.com/MesserLab/SLiM/blob/0627dcacfda8973aacde0c4ad1a3eb4b232eac51/core/slim_sim.cpp#L5304), so a sort of the tables as we go along ought to be OK in princple.  

Furthermore, I've just run a quick test, and inserting a call to `sim.treeSeqOutput` in the middle of a simulation does *not* the final output. (It does change the *order* of the *rows* in the final tree sequence, but the tree sequences are equivalent!)

And, let's see - after sorting and deduplicating sites, Simplify does
1. [simplify](https://github.com/MesserLab/SLiM/blob/0627dcacfda8973aacde0c4ad1a3eb4b232eac51/core/slim_sim.cpp#L5321)
2. [update map of remembered genomes](https://github.com/MesserLab/SLiM/blob/0627dcacfda8973aacde0c4ad1a3eb4b232eac51/core/slim_sim.cpp#L5326) which maps genomes to nodes
3. [record table position](https://github.com/MesserLab/SLiM/blob/0627dcacfda8973aacde0c4ad1a3eb4b232eac51/core/slim_sim.cpp#L5329) 

Skipping step (1) means the number of rows in the tables won't change, so step (3) is unnecessary. And, [tsk_table_collection_sort doesn't affect the node table](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TableCollection.sort), so step (2) should be unnecessary also.

For completeness, here's the test that inserting a call to outputTreeSeq in the middle doesn't change anything:
with `recipe_nonWFa.slim`:
```
initialize()
{
    setSeed(23);
    initializeSLiMModelType("nonWF");
    initializeTreeSeq();
    initializeMutationRate(1e-2);
    initializeMutationType("m1", 0.5, "f", -0.1);
    initializeGenomicElement(g1, 0, 999);
    defineConstant("K", 100);
}

reproduction() {
    subpop.addCrossed(individual, subpop.sampleIndividuals(1));
}

1 early() {
    sim.addSubpop("p1", K);
}

early() {
    p1.fitnessScaling = K / p1.individualCount;
}

10 late() {
    sim.treeSeqOutput("a_gen10.trees");
   sim.simulationFinished();
}
```
and `recipe_nonWFb.slim`:
```
initialize()
{
    setSeed(23);
    initializeSLiMModelType("nonWF");
    initializeTreeSeq();
    initializeMutationRate(1e-2);
    initializeMutationType("m1", 0.5, "f", -0.1);
    initializeGenomicElement(g1, 0, 999);
    defineConstant("K", 100);
}

reproduction() {
    subpop.addCrossed(individual, subpop.sampleIndividuals(1));
}

1 early() {
    sim.addSubpop("p1", K);
}

early() {
    p1.fitnessScaling = K / p1.individualCount;
}

5 early() {
    sim.treeSeqOutput("b_gen5.trees");
}

10 late() {
    sim.treeSeqOutput("b_gen10.trees");
   sim.simulationFinished();
}
```
the only differences being
```
$ diff recipe_nonWFa.slim recipe_nonWFb.slim
26a27,30
> 5 early() {
>     sim.treeSeqOutput("b_gen5.trees");
> }
> 
28c32
<     sim.treeSeqOutput("a_gen10.trees");
---
>     sim.treeSeqOutput("b_gen10.trees");
```
and then running `python3 test_unchanged.py` where `test_unchanged.py` is
```
import tskit

ats = tskit.load("a_gen10.trees")
bts = tskit.load("b_gen10.trees")

assert ats.num_nodes == bts.num_nodes 
assert set([n.metadata['slim_id'] for n in ats.nodes()]) == set([n.metadata['slim_id'] for n in bts.nodes()])

bmap = {}
for j, n in enumerate(bts.nodes()):
    bmap[n.metadata['slim_id']] = j

at = ats.tables
bt = bts.tables

for n in at.nodes:
  m = bmap[n.metadata['slim_id']]
  assert bt.nodes[m] == n

node_map = [bmap[n.metadata['slim_id']] for n in at.nodes]

bt.subset(nodes=node_map, remove_unreferenced=False, reorder_populations=False)
bt.sort()

at.assert_equals(bt, ignore_provenance=True)
```
This last script is a bit complex because it has to reorder rows in the result to get them to match, but once this is done, then they do indeed match.

So - I think this is not a bug. However, moving the sort and deduplicate calls to after the table copy would be harmless, would be less confusing and possibly avoid other errors in the future.

bhaller:
fixes #209

@petrelharp this is what you had in mind, right?

### DIFF
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -7284,30 +7284,33 @@ void SLiMSim::WriteTreeSequence(std::string &p_recording_tree_path, bool p_binar
 	
 	// First we simplify, on the original table collection; we considered doing this on the copy,
 	// but then the copy takes longer and the simplify's work is lost, and there doesn't seem to
-	// be a compelling case for leaving the original tables unsimplified.
+	// be a compelling case for leaving the original tables unsimplified.  Note that Peter has done
+	// a check that calling treeSeqOutput() in the middle of a run does not change the result, although
+	// it *does* change the order of the rows; see https://github.com/MesserLab/SLiM/issues/209
 	if (p_simplify)
 	{
 		SimplifyTreeSequence();
-	}
-	else
-	{
-		// this is done by SimplifyTreeSequence() but we need to do in any case
-		int flags = TSK_NO_CHECK_INTEGRITY;
-#if DEBUG
-		flags = 0;
-#endif
-		ret = tsk_table_collection_sort(&tables_, /* edge_start */ NULL, /* flags */ flags);
-		if (ret < 0) handle_error("tsk_table_collection_sort", ret);
-		
-		// Remove redundant sites we added
-		ret = tsk_table_collection_deduplicate_sites(&tables_, 0);
-		if (ret < 0) handle_error("tsk_table_collection_deduplicate_sites", ret);
 	}
 	
 	// Copy the table collection so that modifications we do for writing don't affect the original tables
 	tsk_table_collection_t output_tables;
 	ret = tsk_table_collection_copy(&tables_, &output_tables, 0);
 	if (ret < 0) handle_error("tsk_table_collection_copy", ret);
+	
+	// Sort and deduplicate; we don't need to do this if we simplified above, since simplification does these steps
+	if (!p_simplify)
+	{
+		int flags = TSK_NO_CHECK_INTEGRITY;
+#if DEBUG
+		flags = 0;
+#endif
+		ret = tsk_table_collection_sort(&output_tables, /* edge_start */ NULL, /* flags */ flags);
+		if (ret < 0) handle_error("tsk_table_collection_sort", ret);
+		
+		// Remove redundant sites we added
+		ret = tsk_table_collection_deduplicate_sites(&output_tables, 0);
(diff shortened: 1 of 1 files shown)


petrelharp:
Yep, that's exactly what I had in mind!

PR #211, opened 2021-09-18
more consistency checks for tables

petrelharp:
Here's a PR which includes https://github.com/tskit-dev/tskit/pull/1722, which turns on testing for
-    mutation time younger than node
-    mixed known/unknown times at a site
-    site of parent mutation agrees with site

by default in `tsk_table_collection_check_integrity`, which is [called by `CheckTreeSeqIntegrity`](https://github.com/MesserLab/SLiM/blob/ea9bb14eeeae5f98b548bf261870f65c6e0a81bb/core/slim_sim.cpp#L7633), and occurs
- [every generation in DEBUG mode](https://github.com/MesserLab/SLiM/blob/ea9bb14eeeae5f98b548bf261870f65c6e0a81bb/core/slim_sim.cpp#L3696) at [various times](https://github.com/MesserLab/SLiM/blob/ea9bb14eeeae5f98b548bf261870f65c6e0a81bb/core/slim_sim.cpp#L4270)
- in [simplification](https://github.com/MesserLab/SLiM/blob/ea9bb14eeeae5f98b548bf261870f65c6e0a81bb/core/slim_sim.cpp#L5285), again in DEBUG mode only

So - I think this should only affect runtime in DEBUG mode (and then not very much)), and I also think that the only thing it might catch is actual errors, because it's only checking for logical consistency errors, not sortedness stuff that we are not bothering with. But it'd be nice to check.

And - I don't think this PR should be merged, I guess, since it's cleaner to just merge in tskit at releases?

### DIFF
--- a/treerec/tskit/core.c
+++ b/treerec/tskit/core.c
@@ -308,6 +308,11 @@ tsk_strerror_internal(int err)
                   "values for any single site.";
             break;
 
+        /* Migration errors */
+        case TSK_ERR_UNSORTED_MIGRATIONS:
+            ret = "Migrations must be sorted by time.";
+            break;
+
         /* Sample errors */
         case TSK_ERR_DUPLICATE_SAMPLE:
             ret = "Duplicate sample value";
@@ -468,8 +473,8 @@ tsk_strerror_internal(int err)
             break;
 
         /* IBD errors */
-        case TSK_ERR_NO_SAMPLE_PAIRS:
-            ret = "There are no possible sample pairs.";
+        case TSK_ERR_SAME_NODES_IN_PAIR:
+            ret = "Both nodes in the sample pair are the same";
             break;
 
         case TSK_ERR_DUPLICATE_SAMPLE_PAIRS:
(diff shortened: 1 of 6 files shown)


bhaller:
OK, I'm running tests on this now.  I think I would like to merge it in (assuming it passes all tests), immediately after tskit merges the main PR in, actually.  I know that's contrary to our usual mode of operation, but I really want these updated consistency checks in place while I'm working on the fix to implement null genomes more broadly; it will give me some peace of mind.  Will that cause any problems for you later on?  And are you aware of any specific dangers to doing so?  Does this PR include only the changes for "moar flags", or does it include other tskit changes made since the last release?

petrelharp:
Sure, that's fine, and *probably* won't cause headaches - this includes other changes since the last tskit release, but we can deal. I should update the treerec README to reflect what's happened here, though.

bhaller:
Sounds good.  Tests are chugging along; so far, so good.

bhaller:
This passed all checks on my side, and did correctly raise an exception in the DEBUG consistency check for Elissa's model even with coalescence checking turned off.  Good stuff, merging now.  Looks like the tskit folks are going to do a handful of cosmetic changes to this before they accept it on their side; feel free to do another PR that brings up exactly in sync with them once that happens, but my impression is that it doesn't matter and we'll get those changes when we next merge anyway.  (Let's watch that they don't change anything significant in this PR, though.  :->)  Thanks!

PR #219, opened 2021-09-25
Windows build

rdinnager:
Compilation of all units works on my Windows machine. 

To Do:

- [x] Fix help file problems (encoding or line-endings most likely)
- [ ] Fix fonts to match other platforms better
- [ ] Fix temporary folder issue
- [ ] Try and find a fix for `Eidos_GetMaxRSS()`
- [ ] Setup test compilation on Windows github actions
- [ ] Make MSYS2 MINGW package that can be installed with `pacman` (https://www.msys2.org/wiki/Creating-Packages/)
- [ ] Look into `conda` packaging
- [ ] Write installation instructions for SLiM manual
- [ ] Benchmark Windows compilation?
- [ ] Add more here as they come up...


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
rdinnager, line 1: Oops, I think I committed this file by accident. I think it is just some setting from the VSCode project I am using. Will remove in next commit.

CMakeLists.txt
```diff
@@ -70,6 +70,14 @@ option(BUILD_SLIMGUI "Build the Qt5-based GUI for SLiM" OFF)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wmissing-prototypes -Wswitch -Wpointer-sign -Wsign-compare -Wstrict-prototypes -Wno-sign-conversion -Wuninitialized")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wswitch -Wsign-compare -Wno-sign-conversion -Wuninitialized")
 
+# Windows specific flags
+if(WIN32)
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fpermissive -D_POSIX_THREAD_SAFE_FUNCTIONS -DDO_MEMORY_CHECKS=0")
```
rdinnager, line 76: I added -fpermissive to convert some errors to warnings, but I realize now that I fixed those warnings later, so we shouldn't need this anymore.

bhaller:
OK, so my initial comment is: you've added includes of "config.h" to everything, making this PR touch 256 files.  I *think* there should be one way to tell the build system on Windows "always include this file at the start of every file you compile".  If you can't find a way to do that, then so be it, but it'd be great to have this PR touch far fewer files.  You're also touching a bunch of files by deleting a single blank line in them; please don't do that, I like have two blank lines between the include protection stuff and the start of the includes.  Basically: please try to prune this PR down to touching the minimal set of files possible.  :->

rdinnager:
Yes, I realized later that I can probably add the config.h header using a preprocessor flag (`-include`), which is what I ended up doing to deal with the automoc'ed files since I couldn't edit them directly. I will do that now that I know everything works. Also, sorry about the deleted space, I added something then deleted, then forgot how many newlines there were. Should have just reverted it. I will fix this too.

rdinnager:
Just leaving a note about how to build this here, in case anyone wants to try it. You need to have MSYS2 installed first (https://www.msys2.org/). Open up the MINGW64 shell (there is also a MINGW32 shell, I haven't tested on this yet, but generally I am assuming most people will want a 64-bit version). Note, do not open the plain MSYS2 shell, it is not set up to use the Mingw compiler, and so this won't work there. In the MINGW64 shell, run these commands from inside a build directory in the SLiM folder:
```
cmake -G "MSYS Makefiles" -DBUILD_SLIMGUI=ON ..
make
```

This will build the executable in the build directory. I haven't set up any install scripts yet. To run SLiMgui.exe, you might have to add the folder that holds the Qt5 .dlls to your Path (usually something like `C:\msys64\mingw64\bin`). To speed things up you might want to add the -j flag to make to parallelize the build.

bhaller:
I see a commit named "got rid of most config.h includes", and I see that its diffs do appear to get rid of most config.h includes, but this PR still says 232 files are changed, and the diffs for this PR still show all the config.h includes.  Not sure what's going on; is git confused somehow?

rdinnager:
Well, I still have to remove `config.h` includes from the `QtSLiM` directory, which I had forgotten about. But most of the 232 files are not changes to existing files, but committing whole new files. I added a new subdirectory called `windows_compat`, which contains the gnulib project files. I assume you are only concerned about the number of files in the original SLiM project that gets changed, but are okay with adding new Windows specific files? If adding the `gnulib` files to `windows_compat` is a problem, please let me know, and I will stop this PR. I don't think the port is very doable without it though.. The `windows_compat` folder is around 170 extra files.

bhaller:
> Well, I still have to remove `config.h` includes from the `QtSLiM` directory, which I had forgotten about. But most of the 232 files are not changes to existing files, but committing whole new files. I added a new subdirectory called `windows_compat`, which contains the gnulib project files. I assume you are only concerned about the number of files in the original SLiM project that gets changed, but are okay with adding new Windows specific files? If adding the `gnulib` files to `windows_compat` is a problem, please let me know, and I will stop this PR. I don't think the port is very doable without it though.. The `windows_compat` folder is around 170 extra files.

Ah, I see.  I had thought that most of the 232 files were config.h touches.  No problem with adding the `windows_compat` folder; maybe name it `windows_gnulib` for clarity?  Or if you plan to put other windows files in there as well, then maybe put the gnulib files in their own subdirectory inside `windows_compat`?  It'd just be good to have the gnulib files separated in their own clear location, so that if they need to be diffed/updated at some point that is easy to do.

bhaller:
I'm just about done with the other thing that's been taking up my time; I expect to commit it tomorrow.  Let me know when this PR is clean and ready for me to have a look.  Thanks!

rdinnager:
I think it is ready for you to look at it now. I think there is one file still where I accidentally inserted an extra blank line, but I can fix that at some point down the line. I still have to try and fix up the font size issues. One problem is that I don't have access to the linux or Mac version of SLiMgui, so I don't know what it is supposed to look like. But I think there are some screenshots in the manual which I can probably use as a reference.

.vscode/settings.json
bhaller: don't forget to remove this...

.gitignore
```diff
@@ -253,6 +253,10 @@ ui_*.h
 *.jsc
 Makefile*
 *build-*
+!build-aux
```
bhaller, line 256: The stuff under `windows_compat` makes sense to me, but it's not usual to need to commit build files; what is this `build-aux` directory and should it really be included?

CMakeLists.txt
```diff
@@ -55,7 +55,7 @@
 cmake_minimum_required (VERSION 2.8 FATAL_ERROR)
 
 project(SLiM)
-
+include(ExternalProject)
```
bhaller, line 58: Does this have any side effects worth worrying about?  Can it be made to happen only on Windows?

CMakeLists.txt
```diff
@@ -70,6 +70,14 @@ option(BUILD_SLIMGUI "Build the Qt5-based GUI for SLiM" OFF)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wmissing-prototypes -Wswitch -Wpointer-sign -Wsign-compare -Wstrict-prototypes -Wno-sign-conversion -Wuninitialized")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wswitch -Wsign-compare -Wno-sign-conversion -Wuninitialized")
 
+# Windows specific flags
+if(WIN32)
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_POSIX_THREAD_SAFE_FUNCTIONS -DDO_MEMORY_CHECKS=0")
```
bhaller, line 76: I'm curious why you add `POSIX_THREAD_SAFE_FUNCTIONS` here.  SLiMgui is single-threaded, so I wouldn't expect this to be needed (and it might slow things down, since it probably introduces threading locks on functions or some such).

CMakeLists.txt
```diff
@@ -152,6 +195,9 @@ find_package(Qt5 REQUIRED
     Gui
     Widgets
 )
+if(WIN32)
+set_source_files_properties("${CMAKE_CURRENT_BINARY_DIR}/${TARGET_NAME}_autogen/mocs_compilation.cpp" PROPERTIES COMPILE_FLAGS "-include config.h -DGNULIB_NAMESPACE=gnulib")
```
bhaller, line 199: Normally SLiMgui is built with qmake, not cmake, which handles the MOC stuff and might handle other issues you're dealing with here.  But hey, if cmake works for you, no worries.

EidosScribe/EidosTextView.mm
```diff
@@ -2001,16 +2001,22 @@ - (NSMutableArray *)completionsForKeyPathEndingInTokenIndex:(int)lastDotTokenInd
 	
 	// First, a sorted list of globals
 	for (auto symbol_sig : *terminus->Properties())
-		[candidates addObject:[NSString stringWithUTF8String:symbol_sig->property_name_.c_str()]];
+	{
+		if (!symbol_sig->deprecated_)
```
bhaller, line 2005: These diffs should not be coming through in your PR.  I think there is a problem with your rebasing.  These are changes that I committed.  I'm going to stop here until this is resolved.

bhaller:
I wonder if maybe it would be a good idea to do two separate PRs: one that adds all of the `windows_compat/gnulib` files, without any other changes, just to get all of that cruft into the project – I could approve that PR immediately – and one that does everything else.  We'll probably be doing multiple rounds of review, etc., on this, and having all the gnulib files in the PR just makes reviewing it difficult.  Let's clear the decks; what do you think?  It'll be quite harmless to approve it, since without other changes that whole directory won't even be used by the build at all; and if for some reason we realize we don't want them after all, it's easy enough to just remove the whole directory again.

rdinnager:
> 
> 
> I wonder if maybe it would be a good idea to do two separate PRs: one that adds all of the `windows_compat/gnulib` files, without any other changes, just to get all of that cruft into the project – I could approve that PR immediately – and one that does everything else. We'll probably be doing multiple rounds of review, etc., on this, and having all the gnulib files in the PR just makes reviewing it difficult. Let's clear the decks; what do you think? It'll be quite harmless to approve it, since without other changes that whole directory won't even be used by the build at all; and if for some reason we realize we don't want them after all, it's easy enough to just remove the whole directory again.

Sure, let's do that first.

EidosScribe/EidosTextView.mm
rdinnager: It is because those commits are absent from the `windows_port` branch, which is what I am PRing into. You will have to merge `master` into `windows_port` (or just delete the branch and then regenerate it so it is even with `master`. Or just leave it and let the commits be added from this PR.. This might lead to a bit more of a confusing commit history though, I'm not sure. I'm also not exactly a master of git..

bhaller: Ah!  Right you are.  I have just merged master's changes into windows_port.  Let me know when things are sorted out; let's do the separate PR of the gnulib files first, as discussed below.  Thanks!

CMakeLists.txt
rdinnager: I can make it happen only on Windows to be safe.

.gitignore
bhaller: I see; so it's not build products, but things needed to build.  Can it go inside the windows_compat folder?

CMakeLists.txt
bhaller: Yes, something like that would be better.  Or you might be able to track down a header that provides `localtime_r` even when this -D is not defined.

.vscode/settings.json
rdinnager: Ha! I did remove it, but I think I accidentally recommitted it again :P. Oh well, I will add it to `.gitignore` so I don't do it again.

.gitignore
rdinnager: It is inside `windows_compat/gnulib`. I assume it is okay there? It is placed there by the `gnulib-tool` which is used to setup `gnulib`, I wouldn't want to change it because this would require manually editing the generated `configure` script, which would be hard to keep track of.

CMakeLists.txt
rdinnager: I just tested with `gnulib`'s `time_r`, and it works without the `POSIX_THREAD_SAFE_FUNCTIONS` flag now, so that is good.

.gitignore
bhaller: Oh, I see.  OK, just change the exclude here in .gitignore to give a path that's inside windows_compat, then.  My goal is for the .gitignore to be more self-documenting, so that it's clear that this is something excluded because of the Windows build, so I don't get confused by it later on.  :->

rdinnager: Okay fair enough! Will do.

EidosScribe/EidosTextView.mm
rdinnager: I'm not sure why these are still here. Time to make a backup and get into the deep git.

rdinnager: Actually, as a test I tried just opening a new PR and looks like this issue is gone. So it is just this PR which is a bit confused. So I'm just going to abandon this PR, and make a new one with `gnulib` (as discussed), and then another with everything else. I will keep this PR open for now to keep our conversation here open.

bhaller: OK, SGTM.  I've seen this problem before with rebasing pull requests; it seems to either not work well, or be easy to do incorrectly.  @petrelharp may have advice.  Anyway, I will refrain from doing any further commits until we have merged in your gnulib PR, and then life ought to be somewhat simpler.

bhaller:
Looks like comments from commits you merged in are appearing in the discussion above, because the rebase is screwed up.  Ignore them.  :->  @rdinnager, I was expecting you to generate a PR for the gnulib stuff that I would merge, and I'm holding off doing any further commits on my end until that happens; but it hasn't happened and now I'm getting itchy.  :->  Please reply regarding this, otherwise I'll go back to committing stuff; lots to do today!  :->

rdinnager:
Oh, that is weird. I'm just going to close this now. I just came back from lunch and working on the new PR now. Should hopefully come through in the next half hour (just want to be careful to not screw anything up, though I have a backup in the worst case).

rdinnager:
I think you should go ahead and start making commits again. Turns out I need to do a bit more research to understand how to remove changes to the non-windows_compat files from the remote branch, whilst making sure I keep them locally, to be added later. I think I've tracked down the answer (involving splitting into multiple local branches) but I don't want to keep you waiting any longer. I will figure out how to keep even with your additional commits.

bhaller:
> I think you should go ahead and start making commits again. Turns out I need to do a bit more research to understand how to remove changes to the non-windows_compat files from the remote branch, whilst making sure I keep them locally, to be added later. I think I've tracked down the answer (involving splitting into multiple local branches) but I don't want to keep you waiting any longer. I will figure out how to keep even with your additional commits.

Sometimes when git gets confused I end up just creating a new copy of my repo and copying all the source files from the old repo to the new repo.  Git is very powerful, but it is not designed in a user-friendly way at all.  :-<

rdinnager:
Yeah, I think I might do something like that.

PR #220, opened 2021-09-29
added windows_compat directory containing gnulib source files

rdinnager:
Okay, here it is. I just ended up copying the whole directory and restarting from a freshly cloned branch off `master`. It works but unfortunately I lost my commit history, which I was trying to avoid, but oh well. I will just keep the old `windows-build` branch in case I ever want to go back and see my commit message during the process of figuring this all out.

The remaining unstaged changes touch 27 files. Let me know when you are ready for a second pull request to get the rest of the changes.

A couple notes:

I just did a recompile from scratch and now there are no warnings except for the one `tskit` one, so that is good.

I managed to get a `MSYS2` package specification working that can install SLiM automatically on windows using `pacman`. I am using my fork as the source for this, but once these changes get merged into the main repository it should be very quick to modify the package to point to it, and then submit a PR to the MINGW-packages repository. Once this is done, the package should become available within a few days for users, which will mean installation will be a breeze.


bhaller:
OK, merged this.  Ready for the other PR now.  The `pacman` stuff sounds great.  Thanks!

PR #224, opened 2021-09-30
Windows port non gnulib

rdinnager:
Okay, finally had to do another full rebase and it finally looks like it worked properly. Sorry about the delay. I think my issues might have something to do with squashing commits by default when merging upstream, which confuses github. I will have to look into it.

I have tested this with a full recompile. I now get no errors in the debug window in SLiMgui since I fixed the working directory issue. Also, all SLiM tests pass. Some Eidos tests fail because some of them refer to `/tmp/` still, and the ones that use `system` still fail obviously.

To Do:

- [x] Fix help file problems (encoding or line-endings most likely)
- [x] Fix fonts to match other platforms better
- [x] Fix temporary folder issue
- [ ] Try and find a fix for `Eidos_GetMaxRSS()`
- [ ] Setup test compilation on Windows github actions
- [x] Make MSYS2 MINGW package that can be installed with `pacman` (https://www.msys2.org/wiki/Creating-Packages/)
- [ ] Look into `conda` packaging
- [ ] Write installation instructions for SLiM manual
- [ ] Benchmark Windows compilation?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,6 @@ if(WIN32)
     set_target_properties(gnu PROPERTIES IMPORTED_LOCATION ${GNU_STATIC_LIB})
     set_target_properties(gnu PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${GNU_INCLUDES})
     set_target_properties(gnu PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ${GNU_INCLUDES})
-
 endif()
 
 # Test for -flto support
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,11 +72,28 @@ option(BUILD_SLIMGUI "Build the Qt5-based GUI for SLiM" OFF)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wmissing-prototypes -Wswitch -Wpointer-sign -Wsign-compare -Wstrict-prototypes -Wno-sign-conversion -Wuninitialized")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wswitch -Wsign-compare -Wno-sign-conversion -Wuninitialized")
 
-# Windows specific flags
+# Windows specific flags and variables
 if(WIN32)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDO_MEMORY_CHECKS=0")
     set(CMAKE_CXX_STANDARD_LIBRARIES "-static-libgcc -static-libstdc++ -lwsock32 -lws2_32 ${CMAKE_CXX_STANDARD_LIBRARIES}")
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-Bstatic,--whole-archive -lwinpthread -Wl,--no-whole-archive")
+    set(GNULIB_NAMESPACE_SOURCES "${PROJECT_SOURCE_DIR}/core/chromosome.cpp" 
+        "${PROJECT_SOURCE_DIR}/core/genome.cpp" 
+        "${PROJECT_SOURCE_DIR}/core/population.cpp"
+        "${PROJECT_SOURCE_DIR}/core/slim_globals.cpp"
+        "${PROJECT_SOURCE_DIR}/core/slim_sim_eidos.cpp"
+        "${PROJECT_SOURCE_DIR}/core/slim_sim.cpp"
+        "${PROJECT_SOURCE_DIR}/core/subpopulation.cpp",
+        "${PROJECT_SOURCE_DIR}/eidos/eidos_functions.cpp"
+        "${PROJECT_SOURCE_DIR}/eidos/eidos_globals.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMAbout.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMAppDelegate.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMFindRecipe.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMGraphView.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMHaplotypeOptions.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMHelpWindow.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMScriptTextEdit.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMVariableBrowser.cpp"
+        "${PROJECT_SOURCE_DIR}/QtSLiM/QtSLiMWindow.cpp")
 endif()
 
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -g -Og -DDEBUG=1 -DSLIMPROFILING=0")
@@ -179,7 +196,7 @@ add_executable(${TARGET_NAME} ${EIDOS_SOURCES})
 target_include_directories(${TARGET_NAME} PRIVATE ${GSL_INCLUDES} "${PROJECT_SOURCE_DIR}/eidos")
 target_link_libraries(${TARGET_NAME} PUBLIC gsl eidos_zlib tables)
 if(WIN32)
-set_source_files_properties(${EIDOS_SOURCES} PROPERTIES COMPILE_FLAGS "-include config.h")
+    set_source_files_properties(${EIDOS_SOURCES} PROPERTIES COMPILE_FLAGS "-include config.h")
     target_include_directories(${TARGET_NAME} BEFORE PUBLIC ${GNU_DIR})
     target_link_libraries(${TARGET_NAME} PUBLIC gnu)
 endif()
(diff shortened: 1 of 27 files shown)


CMakeLists.txt
```diff
@@ -70,6 +72,30 @@ option(BUILD_SLIMGUI "Build the Qt5-based GUI for SLiM" OFF)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-attributes -Wunused-label -Wimplicit -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wmissing-prototypes -Wswitch -Wpointer-sign -Wsign-compare -Wstrict-prototypes -Wno-sign-conversion -Wuninitialized")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wunused-label -Wunused-variable -Wunused-value -Wno-pragmas -Wempty-body -Wshadow -Wparentheses -Wswitch -Wsign-compare -Wno-sign-conversion -Wuninitialized")
 
+# Windows specific flags and variables
+if(WIN32)
+    set(CMAKE_CXX_STANDARD_LIBRARIES "-static-libgcc -static-libstdc++ -lwsock32 -lws2_32 ${CMAKE_CXX_STANDARD_LIBRARIES}")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-Bstatic,--whole-archive -lwinpthread -Wl,--no-whole-archive")
+    set(GNULIB_NAMESPACE_SOURCES "${PROJECT_SOURCE_DIR}/core/chromosome.cpp"
```
bhaller, line 79: I think I see what you're doing with this `GNULIB_NAMESPACE_SOURCES` business.  It is elegant; the only question I have regards maintenance.  From time to time, I will presumably add something to some source file that will create a dependency on gnulib, and that will break the Windows build until you notice it and fix it.  Is there a reason not to just do the gnulib include and define for *all* source files?  Are there source files where that causes a problem?  Or does it just make the build substantially slower?  If it works OK, it might make our lives simpler down the road.

core/slim_eidos_block.cpp
```diff
@@ -23,6 +23,9 @@
 #include <algorithm>
 #include <vector>
 
+#ifdef _WIN32
+#define strtoq strtoll
```
bhaller, line 27: This should no longer be needed, I changed all the strtoq() to strtoll().

eidos/eidos_globals.cpp
```diff
@@ -17,6 +17,7 @@
 //
 //	You should have received a copy of the GNU General Public License along with Eidos.  If not, see <http://www.gnu.org/licenses/>.
 
+// On Windows enable gnulib namespace to avoid clashed with some class methods
```
bhaller, line 20: this comment can be removed...?

eidos/eidos_globals.cpp
```diff
@@ -1266,7 +1292,16 @@ bool Eidos_CreateDirectory(std::string p_path, std::string* p_error_string)
 // This is /tmp/ (with trailing slash!) on macOS and Linux, but will be elsewhere on Windows.  Should be used instead of /tmp/ everywhere.
 std::string Eidos_TemporaryDirectory(void)
 {
+	#ifdef _WIN32
+	std::string temp_path;
+	char charPath[MAX_PATH];
+	if (GetTempPathA(MAX_PATH, charPath))
+  		temp_path = charPath;
+	std::replace(temp_path.begin(), temp_path.end(), '\\', '/'); // replace windows slashes
```
bhaller, line 1300: huh, interesting; the C filesystem APIs expect forward slashes rather than backslashes, even on Windows, then?  or how does this work?

eidos/eidos_interpreter.cpp
```diff
@@ -31,6 +31,10 @@
 #include <algorithm>
 #include <limits.h>
 
+#ifdef _WIN32
+#define strtoq strtoll
```
bhaller, line 35: again, can be removed, strtoq() is no longer called

bhaller (commented):
OK, looking very close now.  If you clean up the remaining minor issues I will approve this PR, when will make things simpler as far as merging and rebasing and so forth.  Any remaining changes can be done in subsequent PRs.  Does that sound right to you?

CMakeLists.txt
rdinnager: To understand why we should only define this for some sources, it would help to under how `gnulib` works. It works by selectively defining functions that are missing on the target system. This works great in C code but causes a problem in C++ code. This is because if you have a class with a method which happens to have the same name as one of the functions being replaced by `gnulib`, that method gets redefined and you get linker errors. The `GNULIB_NAMESPACE` define works around this by telling `gnulib` to define its replacement functions as `gnulib::function_name`, so it no longer clashes with any C++ class methods. So you put this define in any source that has such a clash. The problem is then, if you use any of the functions defined by `gnulib` in that same source, it won't work anymore because you now have to refer to them as `gnulib::function_name` instead of just `function_name`. So if we put this in all sources, we will likely have to do a lot more replacing of `function_name` with `gnulib::function_name`, and that will be just as annoying if not more annoying to maintain I think, and will lead to having to touch a lot more files potentially. On the other hand it is possible that this situation doesn't come up that often, so it might be worth testing out... I don't know, can we leave it for now? I can try and test alternatives a little down the road once this is all set up? 

In the mean time I can set up a gh actions CI for the Windows build that can warn me if the build gets broken by any update.

eidos/eidos_globals.cpp
rdinnager: Oops, yes.

eidos/eidos_globals.cpp
rdinnager: This is a bit of a workaround. I found that some SLiM tests failed if I didn't do this. Somewhere in the Eidos code the backslashes get interpreted as escape characters, which causes an error. This is the easy solution because the funny thing is that Windows understands forward slashes in paths perfectly fine and converts them to backslashes automatically. So the easiest thing is to just always use forward slashes, because they work on both unix and Windows.

rdinnager:
Okay, all done.

CMakeLists.txt
bhaller: I see.  OK, your solution seems best.  Thanks for the explanation.

eidos/eidos_globals.cpp
bhaller: Aha, interesting.  OK, great.  A comment that effect, here where you do the replace, would be great.

bhaller:
> Okay, all done.

OK.  There's one or two small things that need an edit stemming from the replies I just did.  Ping me once that's done, and then I'll do a final review and merge it.

rdinnager:
@bhaller Okay, I think that is the last thing.

bhaller:
Or if you prefer I can approve this now and you can roll further changes into the next PR, of course; don't need to keep making you do these tiny commits, just aiming for polish.  :->

bhaller:
Merged.  Good stuff!  Thanks for your patience with the reviews, I know I'm a perfectionist.  :->  This seems very releasable in SLiM 3.7.  If the `pacman` install is working well, then the writeup for the manual is about all that's needed, right?  The Eidos_GetMaxRSS() is pretty minor, test compilation with GH Actions would be great but can be set up post-release, and conda can be done later.  I'm not sure when SLiM 3.7 is going to roll; there's a user who I was expecting would want me to do it immediately after committing some of my recent work, for publication reasons, but she has gone walkabout so I'm going to just keep stuffing bug fixes in until she resurfaces.  :->

rdinnager:
Sounds good. There is actually one awkward thing about the `pacman` package stuff. I assume you would want the source for the package to be downloaded from the official release on your website? I will need this to get updated to a version with the windows stuff in it before I can submit the package to the msys2 packages repo, and get it ready for users to use. So that means the package won't be ready when the release comes out, because I have to wait for the release to get the package creation process started, does that make sense? Or I could use the github release tarball (actually I think I will have to because `pacman` only supports tarball archives I think), but this still means I have to wait for a new release. I guess this isn't too much of a big deal. I suppose we could put the package installation instructions in the manual for the release with a note that the package may take a week or so to become available (in the mean time, here is compilation instructions... etc.).

bhaller:
> Sounds good. There is actually one awkward thing about the `pacman` package stuff...

Hmm.  Well, it could be based on the official release package on the SLiM website, or it could be based on the GitHub release.  I don't feel strongly one way or the other.  Right now the source archive contains only the code needed to build command-line SLiM, not SLiMgui, but that is unimportant; it can become a full source archive, I have no problem with that.

So here's a question.  If we base it on GitHub, then presumably it will need to be revised whenever a new version is released – it will refer to a specific release version number?  So then we'd change it to the new release tag, and then have to wait for that to get processed?  Whereas if we base it on the source archive on the SLiM website, would that mean that if that archive is updated to a new version the pacman install automatically updates to install the new code?

If that last bit is true, then I see a path forward.  I could produce a new source archive that has all the Windows stuff, and put that onto the website, with a preliminary version number or something – I could do that now, or as soon as we decide that it's sufficiently polished.  It would go up there without fanfare or announcement, and would not be the final 3.7 code, but the pacman process could then be started.  Whenever the 3.7 release is ready, I'd release it and update the source archive, we'd officially announce that it is now ported to Windows, and people would then automatically get the 3.7 release version if they installed at that point.  Does that work, or would people continue getting the pre-3.7 code for some period of time?  How do other developers handle this, do you think?

rdinnager:
I believe your suggestion would work. However, whether we go with Github or the website I'd still need to update the package because it requires specifying the version of any software package being installed. this version number can then be used as a variable to point to the right file to download. So either way, I'd have to update the package, but in either case, all I would have to do is increment the version number. So whichever way you prefer is fine with me. I guess you suggestion makes the most sense, at least initially, since you won't have to make a full official release to get things started.

bhaller:
I see.  Hmm.  Well, yes, I'd prefer to keep the version history on GitHub relatively clean.  I realize now that I also don't want to put prerelease stuff in the source archive on my website, though – I want that to always provide the sources for the current release version, not something prerelease.  But it sounds like I could just put a source archive at a *different* URL on my website, right?  A URL that is otherwise not public, and is used only for pacman purposes?  That's easy to do.  Maybe it would be good to try turning the crank right now, just to see whether everything goes smoothly, even though we don't intend to say anything public about it yet.  What do you think?  Or we can wait a couple more days, at which point almost all the dust will probably have settled.  Anyway, let me know exactly what you need me to post, and when you think I should do it, and I'll follow your lead.  :->

rdinnager:
Yes, that sounds like it would work. Another possibility is to just have a branch of the SLiM github repository for the latest release. So every time you release a new version, make a branch. The `pacman` package can point there (in the same way it is currently pointing to my forked branch). If you didn't want to accumulate a branch for every release, you could always use the same branch name, just delete the old release's branch each time, and them remake the same branch again with the new release. And you could just make the branch now so we can test the package. Then replace it once the official release comes out? Or have the different URL on your website, whatever seems better to you is fine. But yeah, it would probably be good to test it out before the new release goes live, on way or another.

bhaller:
I think I prefer to have a separate URL on my website for pacman; I'd prefer not to have git involved in this process.  I'll make that archive once a couple of things get sorted out today.

PR #226, opened 2021-10-01
fixed Eidos tests on WIndows

rdinnager:
With this, no Eidos or SLiM tests fail on Windows. It should be noted that the input argument for `system()` fails on Windows because the temporary file can't be opened by `std::ofstream()`. I don't know why. It might take some time to work this one out.  Currently, I just did not add any tests for input.

### DIFF
--- a/eidos/eidos_test_functions_other.cpp
+++ b/eidos/eidos_test_functions_other.cpp
@@ -237,10 +237,13 @@ void _RunFunctionFilesystemTests(std::string temp_path)
 		return;
 	
 	// filesAtPath() – hard to know how to test this!  These tests should be true on Un*x machines, anyway – but might be disallowed by file permissions.
-	EidosAssertScriptSuccess_L("type(filesAtPath('/tmp')) == 'string';", true);
+	EidosAssertScriptSuccess_L("type(filesAtPath(tempdir())) == 'string';", true);
+	// these always fail on Windows and I can't think of any good easy replacement
+	#ifndef _WIN32
 	EidosAssertScriptSuccess_L("type(filesAtPath('/tmp/')) == 'string';", true);
 	EidosAssertScriptSuccess("sum(filesAtPath('/') == 'bin');", gStaticEidosValue_Integer1);
 	EidosAssertScriptSuccess("sum(filesAtPath('/', T) == '/bin');", gStaticEidosValue_Integer1);
+	#endif
 	EidosAssertScriptSuccess_NULL("filesAtPath('foo_is_a_bad_path');");
 	
 	// writeFile()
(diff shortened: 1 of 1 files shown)


bhaller:
Looks good to me, merged.  Then I merged in a new commit from master, so we're in sync again.  I think this branch is now ready for a `pacman` test.  Do you concur, or is there something else you want to roll in first?

rdinnager:
I concur.

bhaller:
OK, the archive has been uploaded to [](http://benhaller.com/slim/SLiM_pacman.tar.gz).  Let me know if you need anything else from me for this.  Once it builds successfully in `pacman`, we should probably start talking about when to merge the windows branch back down into master; seems like we're just about there, but you had a checklist somewhere that is maybe not done yet.  Up to you; I think all of these changes should be safe for the non-Windows build, so I'm fine staging this however you want.

rdinnager:
I'm fine with whenever you want to merge the windows branch in. The checklist was more to remind myself what I wanted to do, not all of them are necessary before merging. I'm happy with the state of it now. But let's wait until `pacman` is sorted.

rdinnager:
Hey @bhaller, the pacman package is failing with the `tar.gz` you linked. It has a bunch of extra files on it (starting with `._`). Looks just like the issue described here (and elsewhere):
https://osxastrotricks.wordpress.com/2010/09/24/extra-files-_-dot-underscore-in-tar-archives/

bhaller:
OK, I'll get to it when I have a moment, thanks.

rdinnager:
Thanks. No rush :).

bhaller:
OK @rdinnager, I believe I ought to have fixed it.  Not sure how to check on my end, since these files are invisible on macOS apparently, but I used the environment variable recommended, so they should no longer be there.  :->

rdinnager:
Just to update you, the new archive worked fine. The msys2 package seems to run fine, though I get warnings I never saw before related to gnulib on clang, which I never tested before. I don't think I need to worry about them, but I will do a few tests just to make sure. I already went ahead and submitted a PR to the msys2 packages repo, but it will have to wait for review, and I don't know how long that generally takes, or whether they might be particular about certain things which might require editing the recipe. We might not be able to get this into 3.7 is this ends up taking too long, but I can go ahead and write up the compilation instructions on Windows, so people can try it out. Another alternative is simply putting the executable files up some to be downloaded. The only problem with this approach though, is that user's will need to have the Qt5 .dll files installed and discoverable. The nice thing about the msys2 package is that Qt5 gets installed automatically at the same time as SLiM.

bhaller:
> Just to update you, the new archive worked fine...

Awesome!  My guess is that it will make 3.7; the person who was pushing for a quick release of it now says plans have changed and we likely have several weeks before we will need to release (although that could change).  So I doubt we'll want/need to put up an executable for download; let's not worry about that for now.  Much better if we can use pacman and have Qt installed automatically, as you say.  A writeup of compilation instructions would be great in any case, because there are always people who need to build the current GitHub head version of SLiM in order to get a bug fix that is blocking them; and then a writeup of how to install with pacman would of course be needed.  Screenshots are good if there are any steps that require navigating dialog boxes and panels.  If you email me the writeups, as plain text or Word, I'll integrate them into the manual and send you a PDF proof for review.  No rush, but of course good to keep rolling along.  :->  Sound good?

rdinnager:
Nevermind, I found the files in the repo. It just wasn't working because the path was hard-coded as a full path instead of a relative path. Yes, maybe down the road I would consider getting into the low-level profiling, out of interest if nothing else, but I don't think I have time for it at the moment. I guess I will have to live with the slowdown for now :(. Or I guess i could use WSL if speed in spatial simulation is what I need.

bhaller:
It would certainly be interesting to figure out the cause of the slowdown, but my guess is it will end up being something that is totally out of our control – a difference in operating system memory allocation behavior or processor cache handling or something like that.

PR #231, opened 2021-10-08
add parents column to .trees output

### LINKED ISSUE #152: Should write our pedigree data to .trees files
Now that https://github.com/tskit-dev/tskit/issues/852 has been fixed, I suppose SLiM ought to convert the pedigree info it has into the corresponding tskit info.  I have the impression that pyslim does this already, though (?).  Does it make sense to move responsibility for this into SLiM in some way, or can pyslim construct the info in the tskit table from the pedigree info that SLiM already provides?  I suppose the less that is done in SLiM the faster forward simulation goes; doing this additional bookkeeping during forward simulation would only be additional overhead.  Thoughts @petrelharp?

bhaller:
Fixes #152.  Here's an attempt at adding a parents column to .trees output.  It also changes the usage pattern for the pedigree id to tskit id hash table from (old) being created internally when individuals are added, temporarily, to (new) being kept permanently, which _greatly_ speeds up models that remember a lot of individuals.  The same sort of pedigree id to tskit id hash table is also created to populate the parents column, so the ability to build and use such hash tables has been generalized and is now used in several spots.

@petrelharp, I'd appreciate a really careful, thorough review of this PR.  I'm munging around with tskit's tables in several spots, and I'm not sure I'm doing it quite right; I'm a bit out of my depth.  Please pay very close attention to that code, such as where the parents column and parents offsets get created and put into the individuals table.  I've run this so far with a couple of test models; now I'll kick off my full test suite, which takes quite a while to run with tree-seq and crosschecks enabled on all the tests.  If you could run your test suite too, that would be great.  It would also be good if you added a couple of new tests that check that the parents column actually gets added now, has sensical contents, etc.; I've done some ad hoc tests of it, but my test suite doesn't really test it thoroughly since my test suite doesn't involve Python or pyslim.  :->

I imagine some changes will be needed in pyslim as a result of this.  The parent pedigree IDs in the metadata should be vended by pyslim, of course.  The file version number has been incremented; the old file version doesn't have the parent pedigree IDs in the individuals metadata, the new file version does.  The metadata schema for the individuals table changed (please proofread what I did, I just typed the new schema into the header file and it might have typos).  Maybe that's it?

Thanks!  This should be a nice feature for folks.

### DIFF
--- a/treerec/tskit/tables.c
+++ b/treerec/tskit/tables.c
@@ -9507,11 +9507,15 @@ tsk_table_collection_check_individual_integrity(
                 goto out;
             }
             /* Check parents are ordered */
+			// BCH 10/8/2021: optimistically removing this check, in the hopes that
+			// https://github.com/tskit-dev/tskit/issues/1774 gets approved.
+			/*
             if (check_individual_ordering && individuals.parents[k] != TSK_NULL
                 && individuals.parents[k] >= (tsk_id_t) j) {
                 ret = TSK_ERR_UNSORTED_INDIVIDUALS;
                 goto out;
             }
+			*/
         }
     }
 out:


bhaller:
Some tests in my test suite are failing; not sure why.  The errors are all some variant of this:

```
_InstantiateSLiMObjectsFromTables tsk_treeseq_init(): Individuals must be provided in an order where children are after their parent individuals
Error on script line 71, character 6:
      sim.readFromPopulationFile(“/tmp/slim_trees_test.trees”);
          ^^^^^^^^^^^^^^^^^^^^^^
```

So there's some issue with the ordering of the individuals table.  So... maybe there's a problem with `ReorderIndividualTable()`?  I wonder whether the way we sort it now needs to be additionally constrained by the parents column or something?  I have frankly lost the thread of how all the individuals table sorting works, so I hope you have insight on this @petrelharp.

bhaller:
> Oh, darn it. In [tskit-dev/tskit#1138](https://github.com/tskit-dev/tskit/issues/1138) we decided to require individuals come after their individual parents..
> 
> One solution would be to back out that requirement in tskit. But we've released it, and it seems useful, so we'd rather not do that.

OK.  Out of curiosity: why is it useful?  Makes implementation of some algorithms simpler, or something?

> Alternatively:
> 
> * maybe we could adjust ``ReorderIndividualTable`so that the Alive individuals come in the correct order, just not the first`n` rows? We've got the `Alive` flag, so we know which ones are alive. This would be possible as long as it's always the case that the position in SLiM's internal state of a parent is always earlier than that of any of their alive children. Is that true?
> 
> * or - mabye better - does the ordering in SLiM's internal state always match the SLiM ID, which we could pull out of metadata? Then we wouldn't have to mess around with reordering the individual table at all!

No, I don't think either of these things is true in SLiM.  The internal order of individuals is contingent upon past operations, and cannot be reconstructed from other state, nor does it reflect that other state.  (For example, when an individual dies it gets removed from the vector of individuals, so to avoid having to shift all other individuals down by one position in the vector, the individual at the _end_ of the vector gets moved back to fill the empty position – thereby breaking any sort of relationship between the ordering and anything else.)

rdinnager:
Yes it was a headache. The underlying issue is discussed here: https://github.com/isaacs/github/issues/750 . There doesn't appear to be any plan to fix it. I think doing a full rebase eventually worked for me, though I tend to be a bit scared of rebasing so that is why I didn't try it initially, instead preferring to merge upstream changes. But merging leads to the issue mentioned in that link, e.g. diffs don't get updated when the base branch of a PR changes. Some people suggested changing the base branch to a random branch, then changing it back, but this actually didn't work for me either. Would love to hear if you happen to know how to deal with this too @petrelharp .

petrelharp:
I'll give a rebase a try!

bhaller:
I'm merging in the diffs from this PR by hand into master, will commit soon.  Sigh.

bhaller:
OK, diffs merged into master by hand (https://github.com/MesserLab/SLiM/commit/e0abc33db1a4442ee8dd6bce6d0b66bb53a62eb1); closing this PR since it will go unmerged.

petrelharp:
The problem probably was the post-checkout hook mentioned in the error message:
```
This repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting .git/hooks/post-checkout.
```

PR #241, opened 2021-11-18
add name and description properties/metadata

### LINKED ISSUE #169: Add `name` and `description` to population metadata

bhaller:
Fixes #169.  Adds `name` and `description` properties to Subpopulation, including the necessary Eidos glue and documentation.  Converts the population table metadata from binary to JSON, and adds those two properties to the population metadata.  Adds up-conversion code that can read old binary metadata and convert it to new JSON metadata at .trees read time.  Adds a new prohibition against using the same population identifier (e.g., `p1`) more than once in a model, to guarantee uniqueness of auto-generated `name` property values.


bhaller:
Not ready for review yet.  To do:

- add uniqueness tests for name
- deal with the text .trees I/O stuff in some way (maybe by removing text input)
- testing (none whatsoever has been done so far)

bhaller:
@petrelharp, this PR is ready to review if you've got time (if not, let me know and I'll just merge it, it seems good to me).  The main things needing review are the new population metadata schema itself, in slim_globals.cpp, and the metadata-related logic in slim_sim.cpp where the bulk of the changes are located; you can ignore all the other files if you like.  It passes all tests on my end.  The tskit tests in CI are still broken because they still run against the release tskit that expects the individuals table to be ordered; it'd be good to run them locally against the current tskit that no longer requires that, if that's easy for you to do.

If you've got a sec to teach me a new Git trick, I'd like to squash all the commits in this PR together into one before I merge it into master.  Is that straightforward?  Thanks!

core/slim_globals.cpp
```diff
@@ -1421,9 +1425,13 @@ R"V0G0N({"$schema":"http://json-schema.org/schema#","additionalProperties":false
...
..._fraction":{"description":"The frequency with which females in this subpopulation reproduce clonally (for WF models).","type":"number"},"male_cloning_fraction":{"description":"The frequency with which males in this subpopulation reproduce clonally (for WF models).","type":"number"},"migration_records":{"items":{"properties":{"migration_rate":{"description":"The fraction of children in this subpopulation that are composed of 'migrants' from the source subpopulation (in WF models).","type":"number"},"source_subpop":{"description":"The ID of the subpopulation migrants come from (in WF models).","type":"integer"}},"required":["source_subpop","migration_rate"],"type":"object"},"type":"array"},"name":{"description":"A human-readable name for this subpopulation.","type":"string"},"selfing_fraction":{"description":"The frequency with which individuals in this subpopulation self (for WF models).","type":"number"},"sex_ratio":{"description":"This subpopulation's sex ratio (for WF models).","type":"number"},"slim_id":{"description":"The ID of this population in SLiM. Note that this is called a 'subpopulation' in SLiM.","type":"integer"},"required":["slim_id"],"type":["object","null"]})V0G0N";
```
petrelharp, line 1433: Do we need to make `slim_id` required? What if we just ignored populations without a `slim_id` key?

core/slim_sim.cpp
```diff
@@ -8443,12 +8300,140 @@ void SLiMSim::__ConfigureSubpopulationsFromTables(EidosInterpreter *p_interprete
...
 		
-		SubpopulationMetadataRec *metadata = (SubpopulationMetadataRec *)metadata_char;
-		slim_objectid_t subpop_id = metadata->subpopulation_id_;
+		if (!subpop_metadata.is_object())
+			EIDOS_TERMINATION << "ERROR (SLiMSim::__ConfigureSubpopulationsFromTables): population metadata does not parse as a JSON object; this file cannot be read." << EidosTerminate(nullptr);
+		
+		slim_objectid_t subpop_id;
+		
+		if (!subpop_metadata.contains("slim_id"))
+			EIDOS_TERMINATION << "ERROR (SLiMSim::__ConfigureSubpopulationsFromTables): required population metadata key 'slim_id' is missing; this file cannot be read." << EidosTerminate(nullptr);
```
petrelharp, line 8318: So I guess to make `slim_id` optional you'd have to at this point skip this entry. I'm not sure how annoying that would be.

petrelharp:
Ok - a couple of comments, no worries otherwise!

To squash those commits, you could have a look at the [stdpopsim writeup](https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#rebasing) of this topic, but the tldr; is:
```
git fetch orgin # or, upstream or whatever you call the central repository
git rebase -i origin/master
# change the first letter in all rows except the first one to be `s` instead of `p`
# and if everything looks good
git status # to make sure it's done rebasing
git push -f # force-push to update this PR
```

core/slim_globals.cpp
bhaller: Well, populations that have no metadata at all do get ignored on the assumption that they are "dummies"; see slim_sim.cpp:8296.  We could do that for pops with metadata but without a `slim_id` key, but that seems kinda weird to me; such pops are probably *not* "dummies", so it would probably be surprising to the user if the file read in without errors but those pops just... weren't there.  I guess we're requiring a top-level SLiM metadata entry with a file version it in, etc., so this would not happen if someone just tried to import a .trees directly from msprime into SLiM; they would still (I think?) get an error that the file was not in SLIM format.  Given that, it's probably kinda moot, unless there's a desire/need to have pyslim write out pops that have metadata but do not have a `slim_id`.  So if it's moot, I'd prefer to keep it the way it is here, since just ignoring pops that do not look like "dummies" but that we don't understand seems wrong – not what the user probably wants, at best, and a confusing cause of bugs, at worst.  Is there an argument in favor of it?

core/slim_sim.cpp
bhaller: See comment previous.

bhaller:
Replied to your comments; just the one question to resolve, and if you really feel strongly about it then I will defer to you (but I'd like to understand why).  Oh, and, @petrelharp: can you please run the tskit test suite for SLiM on your side, since it errors out in the GitHub Actions CI right now?  If it's a hassle then never mind, but I'm guessing you're all set up to do it very easily...?  Thanks!

bhaller:
This PR is currently exhibiting a segfault in GitHub Actions during the SLiM self-test:

[https://github.com/MesserLab/SLiM/runs/4274541479?check_suite_focus=true](https://github.com/MesserLab/SLiM/runs/4274541479?check_suite_focus=true)

After some googling, it looks like GitHub Actions has a problem with spurious segfaulting during tests; various people are complaining.  However, there's no way to tell from the logs whether we're being bitten by that, or whether this PR does indeed have a problem that is causing a segfault.  This segfault appears to occur on Linux but not macOS.  @petrelharp if you could run `slim -ts` to do the SLiM self-test, ideally several times, that would be helpful.  If you *do* get a segfault, I have no idea how debugging on Linux works, but if you can get a backtrace for the segfault (perhaps by running in under gdb or lldb?) that would be very useful.  I'll be doing this on Ubuntu on my end, but it'd be good to get a second data point – if I can't get a segfault on Ubuntu running under VirtualBox on a Mac, I'm not sure I will find that 100% reassuring.  :-O

I strongly suspect that the problem is on GitHub's end, because it would be very unusual (first time ever, maybe?) to have a segfault in SLiM that occurs on one platform and not another – the code is pretty cross-platform.  But there's a first time for everything...

bhaller:
Note that because I had to use Valgrind to find the segfault bug, I've been pushing some other Valgrind-derived fixes into this PR – I wanted to fully validate this branch before merging it, given that there was a segfault introduced.  :-O  The other fixes committed here are unrelated to this PR, though, so the diffs for this PR now contain a bunch of unrelated cruft, sorry.  Running full tests on it now that all the Valgrind issues have been resolved.

bhaller:
OK, passed all tests.  The discussion item in the diffs has not been resolved @petrelharp but I don't think it matters, and I want to get this merged; if you feel strongly about that issue please feel free to pursue it with me.  I don't mean this merge as an end to that discussion, just an end to this PR, because I want to work on some other things that need this PR to be merged in.  :->

core/slim_globals.cpp
petrelharp: Here's the motivation: suppose someone runs an msprime simulations with a bunch of pops, but all the samples are in a single one, say population 5. Then they want to annotate it for SLiM. They might think that they just need to add a `slim_id` to only population 5, because that's the only one SLiM will actually use. So, we'll have to tell them that in fact all populations need a slim_id. This is not a big deal, I guess, since they'll have to be annotating other things anyhow.

bhaller: Hmm, or they could remove the metadata from the other pops that they don't want to show up in SLiM; if there's no metadata, SLiM will then ignore those subpops.  They have a decision to make: do they want those other subpops to be visible in SLiM or not?  And the required key forces them to make that decision, one way or the other.  That seems like a good thing to me...?

petrelharp: Removing metadata isn't a good option, since it removes information about those populations (which are still used in the tree sequence!).

bhaller: Hmm; but if they don't want those subpops to load in SLiM at all, then by definition they don't care about that information, right?  They're prepared to drop those subpops completely; given that, the metadata attached to them doesn't seem like it matters.  Either the user wants them – add a slim_id – or they don't – remove the metadata and they will be dropped.  Is there a problem with that perspective, really?

petrelharp: TBC I don't think this is a big deal. But, for clarity...

Nodes in the tree sequence that aren't part of any extant individuals still have `population` attributes; these nodes will never be read by SLiM and so may refer to populations that SLiM will never think about. If we start SLiM off at 200 years ago, does SLiM need a notion of "Anatolia from 10-15Kya" if that's what one of the "populations" in the tree sequence refers to? We certainly wouldn't want to drop the information.

bhaller: But SLiM *will* drop the information, if we do as you propose.  That's why I'm puzzled.  There are only two choices: load the subpop or don't load the subpop.  To load the subpop, we need a `slim_id`.  If we're _not_ going to load the subpop, then it makes sense to me to make the user "state" that intention explicitly by removing the metadata for the subpop, designating it as, in effect, nonexistent; that metadata would not be used anyway, if the subpop was not loaded into SLiM, right?  So we haven't lost anything that would not have been lost anyway.  The worst of all possible worlds, it seems to me, is to silently drop subpopulations from the model (which, as I understand it, is what you are advocating), simply because they lack a `slim_id`; I can easily see that causing confusion and difficulty.  I feel like we're talking past each other, but I don't see the benefit of what you propose, and the drawbacks seem clear.  What am I missing?

petrelharp: Ah, I see. No, sorry, I was imagining that entries in the population table that SLiM didn't use would be just left as they were. =)

bhaller: I'm not sure I know what you mean.  Perhaps I'm being dense.

petrelharp: I meant that I was being dense, I wasn't thinking properly about how SLiM rewrites the population table.

bhaller: Ah!  OK, then we're done.  :->  Of course if this architecture proves to be a problem in future, we can revisit it.  Something to look forward to.  ;->

PR #248, opened 2021-11-28
tskit C API 1.0 and associated changes

### LINKED ISSUE #247: pull in tskit C API 1.0
This needs to happen as soon as they release it; but I'm thinking that we should actually pull in what they've got now, @petrelharp, to give it a little early testing on our side and make sure there are no issues.  I guess it ought to happen on a branch, just in case, so we don't inadvertently screw up the master branch.

Several issues are waiting for this to happen:

- add a "time unit" of "generations" using the new API they added for that, I guess (no issue open)
- use tskit APIs to read/write the reference sequence instead of going to kastore (#180)
- add a call to get tree-seq metadata from disk (#233)
- check that all tskit tests run cleanly (#232)

I think I will proceed with this unless I hear an objection, @petrelharp.  I'll just copy in the current github head from tskit, on a branch, and see what happens.  :->

bhaller:
Fixes #247 (or will when it's done!).

At present, this merges in a prerelease commit of tskit, just to get the ball rolling and see how things go.  I don't plan to merge this PR in until it has the final release version of tskit C API 1.0; until then, I'll just fix other bugs in this PR that depend upon the new tskit changes (see #247 for links to those bugs).


bhaller:
With this first commit, it builds fine except for one error: "No member named `left_root` in `tsk_tree_t`".  This occurs in method `SLiMSim::CheckCoalescenceAfterSimplification()`, at line slim_sim.cpp:5594.  That code is trying to determine whether our tree sequence has coalesced or not.  Here's the context:

```
	for (; (ret == 1) && fully_coalesced; ret = tsk_tree_next(&t))
	{
#if 0
		// If we didn't keep first-generation lineages, or remember genomes, >1 root would mean not coalesced
		if (t.right_sib[t.left_root] != TSK_NULL)
		{
			fully_coalesced = false;
			break;
		}
#else
		// But we do have retained/remembered nodes in the tree, so we need to be smarter; nodes for the first gen
		// ancestors will always be present, giving >1 root in each tree even when we have coalesced, and the
		// remembered individuals may mean that more than one root node has children, too, even when we have
		// coalesced.  What we need to know is: how many roots are there that have >0 *extant* children?  This
		// is what we use the tracked samples for; they are extant individuals.
		for (tsk_id_t root = t.left_root; root != TSK_NULL; root = t.right_sib[root])
		{
			int64_t num_tracked = t.num_tracked_samples[root];
			
			if ((num_tracked > 0) && (num_tracked < extant_node_count))
			{
				fully_coalesced = false;
				break;
			}
		}
#endif
	}
```

Indeed, `left_root` no longer seems to exist.  I'm guessing this has to do with the changes for the "virtual root"?  I have no idea what the appropriate fix is.  @petrelharp do you know what the deal is here?  I believe you are the author of this coalescence checking code, I have no idea what it's doing.  :->

bhaller:
(I have no idea how the "ui icon tweaks for better quality" commit weaselled its way into this PR; I could swear I pushed that commit before I made this branch.  Bizarre.  Well, it isn't doing any harm hanging out in here until this gets merged...)

bhaller:
Just added a call to `tsk_table_collection_set_time_units()` to set the time unit, on save.  I decided to set it to "tick" rather than "generation", in anticipation of planned changes for multispecies (when the general time unit will, in fact, be a "tick"), and because "generation" is a bit misleading for nonWF models anyway – SLiM should never have used that term, in hindsight, but that ship sailed long ago.

Anyhow, the orthogonal question is whether it should be "tick" or "ticks"; @benjeffery is there a stated preference for this, regarding singular vs. plural?

petrelharp:
It looks like we're just checking whether there's more than one root, so we can use [`tree.get_num_roots()`](https://github.com/tskit-dev/tskit/blob/f2704d2c9949263a92dd7f57d69a5f3ef924dff7/c/tskit/trees.h#L404) instead?, So, I think, instead of
```
		if (t.right_sib[t.left_root] != TSK_NULL)
```
do
```
		if (t.get_num_roots(&t) > 1)
```

bhaller:
> For the first use of left_root...

Yep, those fixes seem good, thanks @petrelharp.

benjeffery:
> @benjeffery is there a stated preference for this, regarding singular vs. plural?

Plural seems to be the way to go, as I've seen "generations" "seconds" etc suggested.

bhaller:
Just an update: the tskit_one_point_oh branch has been updated to tskit 0.99.15, and appropriate fixes have been put in to match the changes done on the tskit side with respect to reading/writing the reference sequence. Changes pushed to GitHub. Seems good so far; running the full test suite now, which takes several hours.

bhaller:
OK, all tests passed, including with crosscheck enabled.  Good to go.  I'm going to merge this branch in, since 0.99.15 is an official tskit release and there seem to be no problems with it.  We'll merge in C API 1.0 in a separate pass when the time comes.

PR #255, opened 2021-12-10
Fix scope of properties in population table metadata schema.

grahamgower:
---
Found looking at msprime issue: https://github.com/tskit-dev/msprime/issues/1947

I didn't check that there aren't similar problems with other schemas. And I didn't test this in any way, not even by looking at the pretty-printed json.


bhaller:
Fixed in my own commit, so closing this PR.  Thanks @grahamgower, this was helpful even though I didn't take it.  :->

petrelharp:
For reference, this is fixed in https://github.com/MesserLab/SLiM/commit/08b234881b154c0b711380a64df9c0a1ef74f53b

PR #257, opened 2021-12-12
read in parent pedigree IDs from file

### LINKED ISSUE #256: reloading from file does not reproduce parent IDs
Here's a MWE: first run this script:
```
initialize()
{
    setSeed(23);
    defineGlobal("TREES_FILE", "out.trees"); 
    initializeTreeSeq();
    initializeMutationRate(1e-2);
    initializeMutationType("m1", 0.5, "f", -0.1);
    initializeGenomicElement(g1, 0, 99);
}

1 early() { 
    sim.addSubpop("p1", 10);
}

10 late() {
	for (ind in p1.individuals)
		catn("Individual " + ind.pedigreeID + " has parents " + paste(ind.pedigreeParentIDs));
    sim.treeSeqOutput(TREES_FILE);
    catn("Done.");
    sim.simulationFinished();
}
```
and then this one:
```
initialize()
{
    setSeed(23);
    defineGlobal("TREES_IN", "out.trees"); 
    defineGlobal("TREES_FILE", "out2.trees"); 
    initializeTreeSeq();
    initializeMutationRate(1e-2);
    initializeMutationType("m1", 0.5, "f", -0.1);
    initializeGenomicElement(g1, 0, 99);
}

1 late() { 
    sim.readFromPopulationFile(TREES_IN);
	for (ind in p1.individuals)
		catn("Individual " + ind.pedigreeID + " has parents " + paste(ind.pedigreeParentIDs));
    sim.treeSeqOutput(TREES_FILE, simplify=F);
}
```

Notice on reloading that the individuals' parent IDs have dissappeared. I think there is no reason they can't be reloaded from the .trees file, so they should be? I think the fix would go in around [here](https://github.com/MesserLab/SLiM/blob/master/core/slim_sim.cpp#L8269), but I'm not sure of how everything there is working.

petrelharp:
Closes #256 (or at least fixes that MWE...)

### DIFF
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -8270,6 +8270,8 @@ void SLiMSim::__CreateSubpopulationsFromTabulation(std::unordered_map<slim_objec
 				individual->SetPedigreeID(pedigree_id);
 				pedigree_id_check.emplace_back(pedigree_id);	// we will test for collisions below
 				gSLiM_next_pedigree_id = std::max(gSLiM_next_pedigree_id, pedigree_id + 1);
+
+				individual->SetParentPedigreeID(subpop_info.pedigreeP1_[tabulation_index], subpop_info.pedigreeP2_[tabulation_index]);
 				
 				uint32_t flags = subpop_info.flags_[tabulation_index];
 				if (flags & SLIM_INDIVIDUAL_METADATA_MIGRATED)
--- a/core/slim_sim.cpp
+++ b/core/slim_sim.cpp
@@ -6623,7 +6623,7 @@ void SLiMSim::WritePopulationTable(tsk_table_collection_t *p_tables)
 			} else {
 				tsk_population_id = tsk_population_table_add_row(
 						&p_tables->populations,
-						NULL, 0);
+						"null", 4);
 				if (tsk_population_id < 0) handle_error("tsk_population_table_add_row", tsk_population_id);
 			}
 			assert(tsk_population_id == last_id_written);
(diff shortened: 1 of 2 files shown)


petrelharp:
This seems to make my formerly-failing pyslim tests pass, FWIW.

petrelharp:
I'm stacking a (possible) fix for #258 on this one.

core/slim_sim.cpp
```diff
@@ -8345,12 +8347,6 @@ void SLiMSim::__ConfigureSubpopulationsFromTables(EidosInterpreter *p_interprete
 		// validate and parse metadata; get metadata values or fall back to default values
 		size_t metadata_length = pop_table.metadata_offset[pop_index + 1] - pop_table.metadata_offset[pop_index];
 		
-		if (metadata_length == 0)
```
bhaller, line 8348: This interacts with the struct-to-JSON translation code in _InstantiateSLiMObjectsFromTables() at slim_sim.cpp:9025; if it sees a zero-length struct metadata chunk, it leaves it zero-length, which then hits this case here and gets treated as empty.  So if you remove this case here (which seems like the right change), then in the struct-to-JSON translation code you will need to write out the 4-byte "null" rather than just leaving it empty.  I will fix that with my own commit after accepting this PR, to keep the ball rolling forward; just wanted you to be aware of the existence of that new translation code.

bhaller:
Good stuff, thanks!  See my one comment above, which I will fix on my end.  Merging this PR now.

PR #266, opened 2021-12-20
fix to python recipes

### LINKED ISSUE #267: Running recipes 18.13 II and III results in an error
... it is `AttributeError: 'Mutation' object has no attribute 'position'`.

petrelharp:
There was one error here (mutation objects don't have a `.position` attribute, sites do), and one API change (it's now `reference_sequence.data` for the actual reference sequence).

Closes #267


bhaller:
Groovy, thanks!

PR #268, opened 2022-01-09
Use standards-compliant printf %e format specifier on Windows.

grahamgower:
This only affects older compilers which link against msvcrt.dll
instead of the newer ucrt.dll.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -190,13 +190,6 @@ jobs:
           qmake ..
           make -j 2
 
-  # Just test that the repository can be cloned on windows.
-  tests-windows-clone:
-    runs-on: windows-2019
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-        
   # test windows-latest compilation
   tests-windows-latest-CLI:
     runs-on: windows-latest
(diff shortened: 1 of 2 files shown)


grahamgower:
Problem identified while getting the windows build onto conda-forge: https://github.com/conda-forge/slim-feedstock/pull/10

rdinnager:
The Windows failures are for a reason unrelated to the code patch. I've seen this before. Not sure why it is cropping up now, but I will look into making a fix (might not happen until tomorrow though).

bhaller:
> The Windows failures are for a reason unrelated to the code patch. I've seen this before. Not sure why it is cropping up now, but I will look into making a fix (might not happen until tomorrow though).

If those failures are unrelated, then I'm happy to approve this PR if it is final.  The Windows failures can be addressed in a separate PR when @rdinnager has time.

grahamgower:
I agree the failure is unrelated. I wanted to fix it here though, just to confirm that the current PR doesn't break with modern mingw. But I guess I don't understand the issue.

grahamgower:
I think this is right now. Annoyingly, some other unrelated issue popped up too (msprime build failure during pyslim installation).

rdinnager:
I think this is fine to merge now. It all looks good to me. Indeed the `pyslim` related issue popping up is annoying. Looks like `msprime` cannot find GSL for some reason. I can look into this tomorrow.

bhaller:
OK, merged.  Thanks @grahamgower and @rdinnager!

PR #276, opened 2022-02-01
Check for empty population metadata

petrelharp:
Here's a (possible) fix for the problem.


core/log_file.cpp
```diff
@@ -597,7 +597,7 @@ EidosValue_SP LogFile::ExecuteInstanceMethod(EidosGlobalStringID p_method_id, co
 
 void LogFile::RaiseForLockedHeader(const std::string &p_caller_name)
 {
-	EIDOS_TERMINATION << "ERROR (" << p_caller_name << "): this LogFile has already emitted its header line, so new dada generators cannot be added." << EidosTerminate(nullptr);
+	EIDOS_TERMINATION << "ERROR (" << p_caller_name << "): this LogFile has already emitted its header line, so new data generators cannot be added." << EidosTerminate(nullptr);
```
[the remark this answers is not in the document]
petrelharp: Yes, sorry, I had already fixed it also but not yet pushed; I've just rebased and this went away.

core/slim_sim.cpp
```diff
@@ -5627,10 +5627,26 @@ bool SLiMSim::SubpopulationIDInUse(slim_objectid_t p_subpop_id)
...
+		if (p_subpop_id < (int)tables_.populations.num_rows) {
+			int ret;
+			tsk_population_t row;
+
+			ret = tsk_population_table_get_row(&tables_.populations, p_subpop_id, &row);
+			if (ret != 0) handle_error("tsk_population_table_get_row", ret);
+			if (row.metadata_length > 0) {
+				// Check the metadata is not "null". It would maybe be better
+				// to parse the metadata, though.
+				if ((row.metadata_length != 4) && !strncmp(row.metadata, "null", 4)) {
```
bhaller, line 5644: shouldn't this be `||` not `&&`?  We want to return true (the id is in use) unless the metadata length == 4 and the value is `null`; the negation of that is length != 4 || value != null.

petrelharp: ack, yes, right!

petrelharp: That's what I had first, but then second-guessed myself...

petrelharp:
FYI our plan in stdpopsim is to use the development version of SLiM for our development work until the release comes out. (which should be fine)

bhaller:
> FYI our plan in stdpopsim is to use the development version of SLiM for our development work until the release comes out. (which should be fine)

OK.  If it reaches the point where a release really needs to happen because it is blocking something, let me know.  But probably best to use the development version for now anyway, rather than rolling a release, because other issues with SLiM 3.7 may arise as work proceeds.

bhaller:
This PR looks fine to me now.  I didn't look at the test code, I assume you have observed that the test is red without the fix and turns green with the fix?  So, is this ready to merge?  It seems like the right fix.

petrelharp:
>  I assume you have observed that the test is red without the fix and turns green with the fix?

I did, yes. However, it's not passing on CI now, I'll figure this out...

petrelharp:
Ok, I fixed up the tests.

Also whoops, I forgot to use `strncmp( )` correctly. =) Fixed now; if the tests here pass, we can merge.

Verification that this does as expected with v3.7 and this branch:
```

$ ../../build_v3.7/slim /home/peter/projects/tskit-dev/SLiM/treerec/tests/test_recipes/test_____multipops.slim

...

#WARNING (Genome_Class::ExecuteMethod_addNewMutation): addNewDrawnMutation should probably not be called from an early() event in a WF model; the added mutation will not influence fitness values during offspring generation.
ERROR (Population::AddSubpopulation): subpopulation p0 has been used already, and cannot be used again (to prevent conflicts).

Error on script line 20, character 8:

    sim.addSubpop("p0", 100);


$ # this version:
$ ../../build/slim /home/peter/projects/tskit-dev/SLiM/treerec/tests/test_recipes/test_____multipops.slim

...

#WARNING (Genome_Class::ExecuteMethod_addNewMutation): addNewDrawnMutation should probably not be called from an early() event in a WF model; the added mutation will not influence fitness values during offspring generation.


```

petrelharp:
It looks like all tests are passing, except the Windows tests are failing at installing msprime, an error some grad students here also encountered. That's separate; I'll open an issue (sigh)...

petrelharp:
So - I think we can merge this.

bhaller:
Thanks for handling this @petrelharp, good stuff!

Issue #277, opened 2022-02-01
Windows tests fail on installing msprime


petrelharp:
See for instance https://github.com/MesserLab/SLiM/runs/5028497404?check_suite_focus=true where we get:
```
Run C:\Miniconda\condabin\conda.bat activate anaconda-client-env
  C:\Miniconda\condabin\conda.bat activate anaconda-client-env
  git clone https://github.com/tskit-dev/pyslim.git
  cd pyslim && pip install .
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    MSYSTEM: MINGW64
Cloning into 'pyslim'...
Processing d:\a\slim\slim\pyslim
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting msprime>=1.0.1
  Downloading msprime-1.1.0.tar.gz (850 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Collecting tskit
  Downloading tskit-0.4.1-cp37-cp37m-win_amd64.whl (394 kB)
Collecting kastore
  Downloading kastore-0.3.1-cp37-cp37m-win_amd64.whl (29 kB)
Collecting numpy
  Downloading numpy-1.21.5-cp37-cp37m-win_amd64.whl (14.0 MB)
Collecting newick>=1.3.0
  Downloading newick-1.3.2-py2.py3-none-any.whl (12 kB)
Collecting demes>=0.2
  Downloading demes-0.2.1-py3-none-any.whl (38 kB)
Collecting svgwrite>=1.1.10
  Downloading svgwrite-1.4.1-py3-none-any.whl (66 kB)
Collecting jsonschema>=3.0.0
  Downloading jsonschema-4.4.0-py3-none-any.whl (72 kB)
Collecting humanize
  Downloading humanize-3.14.0-py3-none-any.whl (98 kB)
Collecting attrs>=20.3.0
  Downloading attrs-21.4.0-py2.py3-none-any.whl (60 kB)
Collecting ruamel.yaml>=0.15.78
  Downloading ruamel.yaml-0.17.20-py3-none-any.whl (109 kB)
[...]

Issue #269, opened 2022-01-09
Occasional `slim -testSLiM` failure


grahamgower:
`slim -testSLiM` usually succeeds, but sometimes fails with:
```
slim -testSLiM
[1806] initialize() { initializeSLiMOptions(keepPedigrees=T); initializeMutationRate(1e-7); initializeMutationType('m1', 0.5, 'f', 0.0); initializeGenomicElementType('g1', m1, 1.0); initializeGenomicElement(g1, 0, 99999); initializeRecombinationRate(1e-8); } 1 { sim.addSubpop('p1', 10); } 5 { if (all(p1.individuals[0].relatedness(p1.individuals[1:9]) < 1.0)) stop(); } : FAILURE : no raise during SLiM execution.

FAILURE count: 1
SUCCESS count: 99284
```

grahamgower:
I did manage to reproduce this just now:
```
$ i=0; while slim -testSLiM; do i=$((i+1)); done

...

[1806] initialize() { initializeSLiMOptions(keepPedigrees=T); initializeMutationRate(1e-7); initializeMutationType('m1', 0.5, 'f', 0.0); initializeGenomicElementType('g1', m1, 1.0); initializeGenomicElement(g1, 0, 99999); initializeRecombinationRate(1e-8); } 1 { sim.addSubpop('p1', 10); } 5 { if (all(p1.individuals[0].relatedness(p1.individuals[1:9]) < 1.0)) stop(); } : FAILURE : no raise during SLiM execution.

FAILURE count: 1
SUCCESS count: 99284

$ echo $i
133
```

bhaller:
The test is incorrect after the fixes made to `relatedness()` in SLiM 3.7, I think.

bhaller:
Sorry this took so long to fix, things are busy over here.  :->  The test was just bad; it used to be correct for the old (buggy) definition of relatedness(), but for the new (hopefully correct) definition it was sometimes violated because the test itself was no longer correct.  I just commented out the offending tests, and I think/hope all is now well.  :->  relatedness() is tested more comprehensively in a different way, outside of the unit tests run inside SLiM.

PR #278, opened 2022-02-01
exploratory looking at windows tests

petrelharp:
Not to be merged (yet?); looking at #277.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,8 +93,7 @@ jobs:
         run: |
           source ~/.profile
           conda activate anaconda-client-env
-          git clone https://github.com/tskit-dev/pyslim.git
-          cd pyslim && pip install .
+          pip install pyslim
 
       - name: Build (Debug)
         run: |
@@ -130,7 +129,7 @@ jobs:
           conda activate anaconda-client-env
           export PATH=$PATH:$PWD/Release
           echo $PATH
-          cd treerec/tests && pytest -xv
+          cd treerec/tests && python -m pytest -xv
 
   GUI:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


petrelharp:
I have no idea why this is not working - I can pip install packages that are then not found, and conda install packages that are then not found.

petrelharp:
Oh. Looks like I just had to clear the conda cache. And it failed *this* time because of #269.

bhaller:
> I have no idea why this is not working - I can pip install packages that are then not found, and conda install packages that are then not found.

Probably it's a different issue, but someone in the Messer Lab just had a very similar problem, which was resolved by these comments:

> type pip3 to see what pip you were using, if it’s not the one in your environment it won’t install to your environment
> type python to see what Python you were using, if it’s not the one in your environment it won’t have pyslim even if pip installed it correctly to the environment
> your env is named py2.7 which implies its default python is Python 2, and pip3 installs packages for Python 3 (it’s in the name, pip3), so even if both of those programs are the right ones from your environment, you won’t be able to find a pip3-installed pyslim from a Python 2 shell
> 
> General tip is to not use Python 2 anymore. It’s discontinued and no one supports it anymore. For some years it existed alongside Python 3, and that generated a lot of confusion. But that time is gone, Python 2 has been fully officially phased out since 2019 I think.
> 
> It’s simple if your defaults are all consistently Python 3, as they should be in any new environment you make:
> conda create -n myenv python=3
> conda activate myenv
> # now both "pip" and "python" should default to the env's Python 3 installation
> pip install pyslim
> Then open python and import pyslim and you should be good.

I imagine you already know all that, but your problem sounds so similar that I thought I ought to mention it.

bhaller:
Sorry about #269, I'll get on that.

petrelharp:
Holy crap, I got this to work! Let me rebase and see if it still works.

bhaller:
> Holy crap, I got this to work! Let me rebase and see if it still works.

Do you understand why it didn't work and now it does?  :->

petrelharp:
> Do you understand why it didn't work and now it does? :->

Yes, I think so - by installing all the prerequisites on Windows using conda, not pip, and bumping the conda cache.

Although now I see it's failing again with the same error, wtf. Um, maybe I'll bump the cache again?

petrelharp:
I think we're good to merge this!

PR #280, opened 2022-02-11
Update installation/binaries table

bryce-carson:
This beautifies the table in the README and also includes the new Ubuntu package as an installation method.


bryce-carson:
Could probably keep the redirect to the Messerlab homepage, instead of directly linking to the .pkg on benhaller.com.

Let me know if that is preferred.

PR #289, opened 2022-02-14
update python recipes; closes #285

### LINKED ISSUE #285: need to modernize pyslim-related APIs in recipes
Recipe 17.10 produces several deprecation warnings; replacement code is needed.  Also, apparently the use of `pyslim.load()` should be switched to `tskit.load()`, maybe just in 17.10, maybe throughout?  Everything seems to run OK for now, apart from warnings, so this is being pended to after 3.7.1 due to schedule pressure.  Clarification on the needed changes would be helpful @petrelharp.  Thanks!

petrelharp:
I think this brings all these up to date.

### DIFF
--- a/QtSLiM/recipes/Recipe 17.10 - Adding a neutral burn-in after simulation with recapitation II.py
+++ b/QtSLiM/recipes/Recipe 17.10 - Adding a neutral burn-in after simulation with recapitation II.py
@@ -1,18 +1,18 @@
 # Keywords: Python, tree-sequence recording, tree sequence recording
 
-import msprime, pyslim
+import msprime, tskit, pyslim
 import numpy as np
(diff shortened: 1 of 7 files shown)


petrelharp:
Failing only due to windows/python hell (purgatory?), it seems.

bhaller:
I guess you only know whether it's hell or purgatory in hindsight, right?  If you ever escape from it, then it turns out it was purgatory.  :->

petrelharp:
... and if you're never sure if you've *actually* escaped or if it's just an temporary illusion?

bhaller:
So... maybe it's really just always hell.  ;->

bhaller:
This looks good, thanks @petrelharp!

PR #291, opened 2022-02-17
Add debian directory

bryce-carson:
This directory contains all of the Debian-specific packaging metadata for SLiM. It is necessary to package SLiM for Debian or Ubuntu to have this folder.
Contributing the work back to upstream is not so that it _can_ be packaged, but so that it is archived in a safe place.
It also is a step towards ever being able to make SLiM a "native" Debian package (that is, in the official repositories), if we ever want to do that.


bhaller:
Hi @bryce-carson.  Can this have a README file in it?  (Maybe it would be good to add one to the Windows folder you added a while back, too, actually, if it doesn't have one already.)  Good for things to be self-documenting to the extent possible; no doubt I will forget later why this folder is here.  :->  Thanks!

bryce-carson:
Hey @bhaller , I can add one. There's already a use for a README.debian, though. I'm not sure if including extra files that aren't part of the guidelines will interfere with anything, but I can definitely write up a little README.debian and include the information you asked for (why this folder exists) in that.

https://www.debian.org/doc/manuals/maint-guide/dother.en.html#readme

bryce-carson:
> Maybe it would be good to add one to the Windows folder you added a while back, too, actually, if it doesn't have one already.

That wasn't me, @bhaller . I think that came from the contributor who got SLiM to compile outside of WSL on Windows.

bhaller:
> > Maybe it would be good to add one to the Windows folder you added a while back, too, actually, if it doesn't have one already.
> 
> That wasn't me, @bhaller . I think that came from the contributor who got SLiM to compile outside of WSL on Windows.

Oh, sorry, yes.  Momentary lapse.  :->

PR #292, opened 2022-02-17
Add RPM spec file

bryce-carson:
This spec file is necessary to package SLiM on Fedora, Red Hat, CentOS, openSUSE, and other RPM-consuming distributions.
It isn't necessary to have the file _in_ SLiM for SLiM to be packaged, but it is a good place to archive the file and also contributes to upstream.


bryce-carson:
For this one, @bhaller , I'm not sure where I'd include a README, but you're a little familiar with the file now.

Perhaps a line or two in the top-level README.md can refer to both this RPM spec file, the debian directory, and also the windows_compat directory (if we can get a word from the contributor of that directory for what it does).

Then it helps some _really advanced_ users who want to rebuild their own packages directly from source. Maybe they're paranoid. Maybe there's a bug. You never know.

Regardless, archival with upstream is great. :)

bhaller:
I'd just put a README.txt file in the directory itself?  I don't think it needs to go in the top-level README, it would just confuse most ordinary mortals.  :->  And I think it's somewhat standard for each subfolder whose purpose is not immediately obvious to have its own README...?

bryce-carson:
> I'd just put a README.txt file in the directory itself? I don't think it needs to go in the top-level README, it would just confuse most ordinary mortals. :-> And I think it's somewhat standard for each subfolder whose purpose is not immediately obvious to have its own README...?

Ah, okay. I proposed that solution because for RPM spec files (a single-file) the norm is to place it at the top-level (that is where Copr would normally look first if you were building directly from GitHub).

This is a new top-level file, however. I'm not sure where I would add a README just for it. I could add extra comments to the top of the file, but if you're already opening the file it should _become_ obvious that it's related to packaging for Fedora and Red Hat.

bhaller:
> ...
> 
> This is a new top-level file, however. I'm not sure where I would add a README just for it. I could add extra comments to the top of the file, but if you're already opening the file it should _become_ obvious that it's related to packaging for Fedora and Red Hat.

Ah, I think I conflated this discussion with the discussion of a README file over in the other PR where you are adding a new directory, for debian I guess.  Yes, I'm not sure where a README for this PR would go.  You know what?  Maybe what we need is a new top-level file, CONTENTS.txt or some such name, that describes what is in the repo – what every file/directory is for, at the top level.  The project is getting big enough that that would probably be helpful.  That doesn't belong in the README.md, because that's more for end-users than for people maintaining the project.  Does that sound like a good idea to you?  If so, I can create it and commit it, and then in your PRs for distro stuff you can add/modify it as needed...?

Sorry for the slow responses on these issues, life is a bit busy right at the moment.  It isn't clear to me whether these PRs are ready to be merged yet or whether you're still working on them; when they're ready to go, just let me know with a comment.  Thanks!  Oh, and I saw your post on slim-discuss, good stuff.  I'll be curious to hear whether you get responses about beta-testing the install packages.  Hopefully the community will recognize the need to contribute to this effort; but people in academia always have too much to do and too little time to do it in...

bryce-carson:
Yeah, a CONTENTS.txt file works and is a good idea.

The PRs are ready now, but once you add a CONTENTS file I'll add new commits to describe the Debian folder and the RPM spec file before you merge the two PRs from me.

bhaller:
> Yeah, a CONTENTS.txt file works and is a good idea.
> 
> The PRs are ready now, but once you add a CONTENTS file I'll add new commits to describe the Debian folder and the RPM spec file before you merge the two PRs from me.

OK.  I've just added CONTENTS.txt.  Give me a ping when the PRs and ready for merging.  Thanks!

bryce-carson:
> > Yeah, a CONTENTS.txt file works and is a good idea.
> > The PRs are ready now, but once you add a CONTENTS file I'll add new commits to describe the Debian folder and the RPM spec file before you merge the two PRs from me.
> 
> OK. I've just added CONTENTS.txt. Give me a ping when the PRs and ready for merging. Thanks!

I described both `debian/` and the RPM spec file in `CONTENTS.txt`. Both PRs are ready for merging if you're fine with the way I have described the files.

I also made a slight change to the description of the desktop files (`org.messerlab.*`).

bhaller:
Looks fine to me, except: should "DISTRIBUTION FILES" and "LINUX PACKAGING" be different categories, as your commit has them, or should they be in the same category?  I'm fine with whatever you want to do, but I'm a bit surprised that you made a new category; I had made "DISTRIBUTION FILES" with the intent that that would fit your purposes.

bryce-carson:
Ah, yes @bhaller. "DISTRIBUTION FILES" works fine, actually. I'll make a new commit and just merge the categories and then the two PRs will be ready.

bhaller:
Great; both merged.  Thanks!

PR #300, opened 2022-03-14
fix multispecies on Windows

### LINKED ISSUE #299: multispecies branch builds failing on Windows
Hi @rdinnager.  This is not urgent, but when you have a chance, I've got a Windows issue I'm unsure how to fix.  You can see the build failures here:

[](https://github.com/MesserLab/SLiM/actions/runs/1971538536)

This is on the `multispecies` branch of SLiM's repo, which is going to be SLiM 4.0 eventually.  The CLI builds fail on Linux too, but _those_ fail in the post-compile tests because of compatibility problems between the new SLiM and the old pyslim.  The Windows build failures, however, seem to be something different, something to do with stringstream and the standard library and such.  I imagine it will be an easy fix for you; I just haven't learned what is needed to keep the Windows build happy when I make changes.  I'll be interested to see what your fix is.  Thanks!

rdinnager:
As you suspected, a pretty easy fix (still fails but it does so on treesequence tests, just like the linux tests). If you see similar errors in the future, they will likely be fixed by adding the problematic source files into the `GNULIB_NAMESPACE_SOURCES` list in `CMakeLists.txt`. Fixes #299 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,7 @@ if(WIN32)
         "${PROJECT_SOURCE_DIR}/core/slim_globals.cpp"
         "${PROJECT_SOURCE_DIR}/core/slim_sim_eidos.cpp"
         "${PROJECT_SOURCE_DIR}/core/slim_sim.cpp"
+        "${PROJECT_SOURCE_DIR}/core/species.cpp"
         "${PROJECT_SOURCE_DIR}/core/subpopulation.cpp"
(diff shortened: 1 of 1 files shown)


bhaller:
Thanks @rdinnager!  Out of curiosity, is there a downside to being listed in `GNULIB_NAMESPACE_SOURCES`?  Is there a reason we don't just put _all_ the source files in that list?

PR #315, opened 2022-05-14
Tskit C_1.0 beta

petrelharp:
So far, it builds! I've done no tests yet. I've also removed vargen, so this closes #304 also.

PR #320, opened 2022-05-21
store name and description in provenance also

petrelharp:
`name` and `description` are now in the metadata; in pyslim I've got some code that reconstructs metadata from provenance, mostly to read in very-old SLiM files, so I ran in to not having those things available. This is not terribly important, but adding them to provenance was (a) the easiest fix and (b) seems like a good idea.


petrelharp:
Oh, whoops - I'll just roll this in to #315, so don't merge this - but have a look at the last commit here, see if you object.

bhaller:
> Oh, whoops - I'll just roll this in to #315, so don't merge this - but have a look at the last commit here, see if you object.

Looks fine to me.  A bit weird that we have so much stuff duplicated between metadata and provenance, but perhaps a case can be made for it being correct, and in any case it doesn't really bother me, if it makes your life easier.  :->

PR #327, opened 2022-06-01
keep slim names in non-used populations

petrelharp:
As discussed, names are an important part of population metadata, which (I think) we would like to therefore retain, independently of whether SLiM is going to use the metadata.

### DIFF
--- a/core/species.cpp
+++ b/core/species.cpp
@@ -4989,7 +4989,8 @@ void Species::WritePopulationTable(tsk_table_collection_t *p_tables)
 		// binary metadata got translated to JSON by _InstantiateSLiMObjectsFromTables() on read
 		while (last_id_written < subpop_id - 1)
 		{
-			bool write_empty_row = true;
+			const char *new_metadata_str = "null";
+			uint32_t new_metadata_length = 4;
 			
 			if (++last_id_written < (slim_objectid_t) population_table_copy->num_rows)
 			{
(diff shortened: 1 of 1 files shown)


core/species.cpp
```diff
@@ -5000,17 +4999,31 @@ void Species::WritePopulationTable(tsk_table_collection_t *p_tables)
...
 					tsk_population_id = tsk_population_table_add_row(
 							&p_tables->populations,
 							tsk_population_object.metadata,
 							tsk_population_object.metadata_length);
 					if (tsk_population_id < 0) handle_error("tsk_population_table_add_row", tsk_population_id);
-					write_empty_row = false;
+				} else {
+					// SLiM metadata for non-extant subpops gets removed at write time for consistency.
+					// See issue #317.
+					// However, we keep names.
```
bhaller, line 5010: Please add an explanation about *why* we keep names, for posterity...

core/species.cpp
```diff
@@ -5000,17 +4999,31 @@ void Species::WritePopulationTable(tsk_table_collection_t *p_tables)
...
+					std::string metadata_string(
+							tsk_population_object.metadata,
+							tsk_population_object.metadata_length);
+					nlohmann::json old_metadata;
+					old_metadata = nlohmann::json::parse(metadata_string);
+					nlohmann::json new_metadata = nlohmann::json::object();
+					new_metadata["name"] = old_metadata["name"];
+					std::string metadata_rec = new_metadata.dump();
+					tsk_population_table_add_row(
```
bhaller, line 5019: add a check of the return value, as above?

bhaller (commented):
Also – GitHub wouldn't let me comment on this line – what about the loop at the end, starting with the comment "// finally, write out empty entries for the rest of the table"?  There could be old SLiM metadata with a "name" value in that area, couldn't there?  Or no?

petrelharp:
I've rebased and squashed this; it's passing my pyslim tests - what do you think?

core/species.cpp
```diff
@@ -4999,24 +5000,44 @@ void Species::WritePopulationTable(tsk_table_collection_t *p_tables)
...
 							tsk_population_object.metadata_length);
-					if (tsk_population_id < 0) handle_error("tsk_population_table_add_row", tsk_population_id);
-					write_empty_row = false;
+					nlohmann::json old_metadata;
+					old_metadata = nlohmann::json::parse(metadata_string);
+					nlohmann::json new_metadata = nlohmann::json::object();
+					if (old_metadata.contains("name")) {
+						new_metadata["name"] = old_metadata["name"];
+						std::string metadata_rec = new_metadata.dump();
+						new_metadata_str = (const char *)metadata_rec.c_str();
```
bhaller, line 5030: Unfortunately for this elegant solution, this is unsafe.  `new_metadata` goes out of scope at the end of the block, and the pointer taken from `c_str()` here becomes a pointer into an object that no longer exists.  If you're lucky the memory stays valid until `new_metadata_str` is used below, but there is no guarantee of that; this is "undefined behavior", which means whether it works or whether it crashes is undefined.  Sorry, needs to be rewritten, probably back to the kind of structure it had before where the call to `tsk_population_table_add_row()` happened within the scope where the new metadata still points to valid data.

bhaller: Hmm, there's a way to rescue the design, I think, by switching from keeping `new_metadata_str` as a C string to keeping it as a std::string (which means the name will get copied, rather than a pointer to the inside of the string object getting taken).  I'm happy to do it.  Give me the green light and I'll accept this PR and fix it on my end.

bhaller:
OK, accepting this and then I've do some more work on it on my side.

core/species.cpp
petrelharp: This actually makes sense!

PR #328, opened 2022-06-02
New tskit1.0

petrelharp:
Here's the final tskit C_1.0 release code. I've kept in your fix to the table output though. It passed all the tests on my machine.


bhaller:
Awesome, thanks!

Issue #334, opened 2022-07-05
Segfault in python-based tests


bhaller:
Hi @petrelharp.  Here's a segfault that occurred in the CI of my latest commit:

https://github.com/MesserLab/SLiM/runs/7205270523?check_suite_focus=true

I'm pretty sure this has nothing to do with the commit; all I did was add a warning in an obscure code path.  So this might be some sort of intermittent problem...?  I can click "re-run" to see if it happens again, but I'm not sure whether that writes over the logs from this run, so I thought I'd call it to your attention before I did that, in case there's something you can ascertain from it.  Let me know if you want me to re-run.  Thanks.

bhaller:
Interesting, it let me assign it to you.  I think it used to not allow me to do that.  :->

petrelharp:
Well, that's vaguely worrisome. here's a screenshot for posterity. I guess the thing to do is to try running that test with a whole bunch of seeds (although I thought the seeds were set!) and see if it happens reproducibly.
![Screenshot from 2022-07-05 19-44-43](https://user-images.githubusercontent.com/1046249/177457273-2e6014d9-6af1-4557-bd2a-ecd7fba4aa51.png)
![Screenshot from 2022-07-05 19-44-34](https://user-images.githubusercontent.com/1046249/177457276-bf2219a7-ef95-4881-b3b8-b60bf69aaadb.png)

petrelharp:
But yes - hit re-run now?

bhaller:
> But yes - hit re-run now?

OK, re-running now:

https://github.com/MesserLab/SLiM/runs/7207688051?check_suite_focus=true

There was a checkbox for "enable debug logs".  I have no idea what it does, but I checked it, because why not.  :->

bhaller:
Looks like the old logs continue to exist; the re-run is separate.  Good to know.

bhaller:
Looks like it segfaulted again @petrelharp, so at least it seems reproducible...?

petrelharp:
Ok! But only on MacOS! Any chance you, as a mac-using-person, could check if this occurs on your machine?

bhaller:
Sure; how do I do so?

petrelharp:
Let's see - briefly; install tskit, pyslim  (version 1.0b1) and pytest; then go to `treerec/tests/` and run `python -m pytest`. As for how to install those - look at the tskit docs, and install pyslim and pytest the same way? (eg with conda?)

bhaller:
[...]

PR #338, opened 2022-07-21
thrash about with CI until I debug #334

petrelharp:


petrelharp:
Well, this just passed the tests, oddly.

petrelharp:
OK, to get the error to occur here, I ssh in and then do:
```
source ~/.profile
set +e
set +o pipefail
export PATH=$PWD/Release:$PATH
cd treerec/tests
python3 -- -m pytest test_consistency.py::TestWithMutations::test_ts_slim_consistency\[test_____sexual_nonwf.slim\]
```
but then I can't get the segfault to occur in a debugger! I've done this a lot of times, no errors!
```
lldb /usr/local/miniconda/envs/anaconda-client-env/bin/python3 -- -m pytest test_consistency.py::TestWithMutations::test_ts_slim_consistency\[test_____sexual_nonwf.slim\]
```

bhaller:
> ...but then I can't get the segfault to occur in a debugger! I've done this a lot of times, no errors!

Ah, always a joy when you get a bug that disappears when you're under the debugger.  :-O  That is probably because running under the debugger changes the layout of memory in some way, or switches to a different version of the standard library (different implementation of malloc, for example), or something like that.

Inserting print statements at strategic points is one strategy then.  You know the line of the test script that triggers the crash; so go inside the code that that line calls, and put print statements all over the place so you can see how far it gets before the crash occurs; that will let you pinpoint the area where the crash is happening.  Then add more print statements inside/before that area to examine all the relevant values.  Figure out what exactly is triggering the segfault – presumably a pointer to a bad address – and then trace backwards from there to find where the bad address comes from.  It can be a nightmare.  :-<

Another strategy, perhaps easier to pursue but it might or might not lead you to the bug, is to run the test, in the specific conditions where it segfaults, under Valgrind.  If you're lucky, Valgrind will lead you right to the specific spot where an uninitialized value is first used.  But if the error is more of a logic error than an undefined behavior error, Valgrind might not tell you very much.  It might still show you the specific line of code where the segfault happens.

petrelharp:
I figured out how to get the bug in a debugger; see https://github.com/tskit-dev/tskit/issues/2429

petrelharp:
Resolved!

PR #340, opened 2022-08-02
Display the SHA-1 for the Git repo SLiM was built from

bhaller:
Trying this out, @bryce-carson.  Probably some loose ends to be tied off.  Thanks!


bhaller:
Wow, it made this based off of master for some reason. WTF.

PR #344, opened 2022-08-11
Rename the RPM spec file, then update it (`SLiM.spec`)

bryce-carson:
This simplifies the conditionals, removing old distribution targets (openSUSE Leap 15.2), as well as serializes and unnests the conditionals to make them more readable and maintainable.


bhaller:
Thanks @bryce-carson.

Issue #288, opened 2022-02-13
[DEPLOYMENT] Debian & Ubuntu Packaging


bryce-carson:
### Motivation
The preferred installation method as of 2022-02-13 is the [installation script](https://github.com/MesserLab/SLiM-Extras/blob/6ea3dbc599ae9f1a225d9b151a7d6961cf0cfea4/installation/DebianUbuntuInstall.sh). DEB packages for Debian and Ubuntu users would be preferred for several reasons, as users can

- Remove the software
- Upgrade the software
- Install the software

all a little bit easier than the script, which also makes users wait for software compilation (and requires compilation software be installed).

### Targets
- [x] Investigate using [**package**cloud](https://packagecloud.io/) for hosting an APT repository for Debian (and potentially Ubuntu as well, forgoing Launchpad)
- [x] Consult the `pbuilder` and `pbuilder-dist` documentation and setup a playground of virtual machines to test building SLiM for all sorts of Debian and Ubuntu versions

### Status: 🚴🏼, slowly.
**Ubuntu**
- [x] 18.04 LTS (Bionic Beaver)
- [x] 20.04 LTS (Focal Fossa)
- [x] 21.10 (Impish Indri)

**Debian**
- [x] 10 ("Buster")
- [x] 11 ("Bullseye")
- [x] Testing ("Bookworm")
- [x] Sid (referred to as "Trixie" on **Package**cloud for some reason)

#### ⚠️ CAUTION ⚠️
The packages for Debian 11 "Bookworm" and Unstable "Sid" were not build with the distribution codename in the package name, but were renamed to include that to permit uploading to **Package**cloud. **Package**cloud disallows packages with the exact same filename to be uploaded, despite different checksums and distribution flags in the `.changes`, etc.

The packages *should* be rebuilt with the distribution codename included in the package name, but difficulties with RAM consumption, necessitating text-mode operation, or using a chroot into the qcow2 image are preventing ease of rebuild.

## Testing
🥇All of the binary packages have been uploaded to either Launchpad or Packagecloud now. The only remaining task is to have all of these sufficiently tested.

Right now that stands at:
**Ubuntu**
- [ ]  18.04 LTS (Bionic Beaver)
  _Maintainer test:
  :warning: Two packages had to be manually installed before `slimsim`: gcc-10-base, and libgcc-s1. These were not available for installation with APT on 18.04 LTS Server after installing X and GNUstep. Hopefully this is not the case for regular desktop Ubuntu.
[...]

Issue #174, opened 2021-03-20
Copr Integration for RPM Packaging CI/CD


bryce-carson:
@bhaller, if we determine how to use the Commitish field to specify *tags*, we can use a Copr webhook.
I don't know how Commitishes work, so how you would specify a tag like I'm not sure, is it a regex like `v[[:digit:]](.[[:digit:]])*`?

> #### SCM
> This method allows you to build RPM(s) from any Git, DistGit, or SVN repository containing a valid .spec file. The only required argument is Clone URL and if the target repository places the .spec file together with package sources in the root directory and you want to build from master HEAD, it will simply work. There are more things to configure for more complex cases. E.g. you might want to specify Subdirectory of the target repository if that is where the repository has the package sources placed. See the following list for the full option description:
> - Type: SCM type of the repository being pointed to by Clone URL (in other words, whether we should use plain git or git svn for subsequent cloning).
> - Clone URL: What repository we should clone to obtain the sources.
> - **Committish:** What tag, branch, or commit we should check out from the history of the cloned repository. By default HEAD of master branch.
> - Subdirectory: Where the subsequent SRPM build command (see below) should be executed. The path is relative to the repository root.
> - Spec File: Path to the spec file relative to the given Subdirectory. Note that you can optionally anchor the path with / (e.g. /rpm/example.spec). If not specified the .spec file will be auto-located.

Using this, we can automatically trigger builds for the HEAD of the master branch, by defining a new package `SLiM-testing` with the SCM package type on Copr. Then the status badge on the README will actually reflect the CD status of the repository.

We can also use this method for future tagged releases.

*This requires a PR of the RPM .spec file to the repository, and for you to configure the web hook on the GitHub repository settings.* You've got administrator access to the Copr repository, so if you want to look at this you should be able to go through the process yourself. If you need my help, however, I am available.

The following files would need to be in the main part of the repository for a simple configuration, but if you'd like you can use a Copr subdirectory and place some files there if you tell Copr where to find them and modify the RPM spec file a little.

- fedora34.patch
- SLiM-3.6.spec

[filesForCoprIntegration.zip](https://github.com/MesserLab/SLiM/files/6176485/filesForCoprIntegration.zip)

bhaller:
Hi Bryce.  I'm afraid I don't really understand what you're talking about here, sorry.  I've never heard of Committish, don't know what a Copr webhook is, etc.  Can you clarify?

bryce-carson:
[...]

PR #346, opened 2022-08-24
Flatpak merge request

### LINKED ISSUE #175: Official Fedora Package: input requested
@bhaller,

I'm going through the process of submitting SLiM to the official Fedora software repository. Unfortunately, another software project (Simple Login Manager) is packaged in Fedora (and other distributions) under the name slim. That project also uses the same capitalization in their project website (SLiM). This causes two issues (that aren't problems in Selection on Linked Mutations itself) in the packaging process:

1. SLiM is not an acceptable package name in Fedora's repositories, as a difference in case alone does not resolve a naming conflict (according to the guidelines).
2. `/usr/bin/slim` is also used by Simple Login Manager, preventing us from using it.

To have SLiM accepted into the Fedora Project's official software repository, there are some solutions proposed by Fedora's Packaging Guidelines:

1. Rename the package to solve the conflict, which I have already done (I have renamed the package to `SLiMgui`; this does not affect users of the Copr repository; installation instructions in the SLiM Manual do not need to change yet).
2. Approach the upstream developers (one or both) and request them to rename the binary to resolve the conflict or to something less conflicting. I have thought of `SLiMcli` as a potential alternative binary name, which parallels the binary name `SLiMgui`.
3. If one or neither upstream developer is unwilling to rename a binary, include a `Conflicts:` tag in the package specification (the .spec file). This prevents users from installing both packages at the same time.

The Fedora Packaging Guidelines state the third solution is not preferable and should be used only as a last resort. This allows users of Fedora to install as many potential packages as possible from the official repositories without requiring manual conflict resolution, which is difficult for the end user.

@iwamatsu's Simple Login Manager (stylized as SLiM) is, as the name implies, a login manager for Linux systems. It's been around for quite a long time, long enough that it hasn't seen any updates since 2013.

In the Fedora Packaging Guidelines, in §[Incompatible Binary Files with Conflicting Naming](https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts/#_incompatible_binary_files_with_conflicting_naming_and_stubborn_upstreams), it's advised that a packager (myself) try to convince one or both upstream projects to rename the binary.

This is not ideal for our case, since:

- Simple Login Manager is not fully compatible with [systemd](https://systemd.io/), a fundamental [and default] part of a modern Fedora Workstation installation. However, @Hubbitus, the Fedora package maintainer of slim, and others contributing to that package have made updates to it in the package itself, so while upstream has abandoned the project it is not stale.
- Simple Login Manager has only a single binary, which would generally not be used by the end-user except in an `~/.xinitrc` file, which launches the login manager upon system boot; end users of Selection on Linked Mutations will occasionally or frequently (depending on how you look at it) use our `/usr/bin/slim` executable to run their models non-interactively.
- The SLiM (Selection on Linked Mutations) Manual has many references to `slim`, which is commonly used by end-users and would unfortunately require changing every reference in the large set of documentation we have.
- Users of the SLiM framework would need to update their scripts, systems, etc. to point to the renamed binary, if we chose to rename it. This would not affect current installations, only future installations; unfortunately, this will cause missing file(s) errors when users have not updated scripts they recieve from others or SLiM-Extras. It would be inadvisable, to say the least, to rename our binary.

The principle binary included in the package, SLiMgui, is the graphical application; we may opt to split the package into two separate packges, one for SLiMgui, and the other for "SLiMcli". In this way, we might kick the conflict down the road for the cli binary of slim, and have SLiMgui accepted into the official Fedora software repository sooner. Necessarily, this would require users who want the command-line program to build and install it from source in the meantime, or continue to use the Copr repository (which would, or could, be deprecated in the future).

I have temporarily renamed the slim binary to SLiMcli, to parallel SLiMgui, in my package definition. This hasn't been and won't be published to Copr. It's a potential solution only at this stage. For end-users, we might advise using `alias slim=SLiMcli` in `.bashrc` or `.bash_profile`; as far as I understand, `getty` is used to bootstrap a graphical Linux environment, so `.bash_profile` isn't executed and wouldn't mask Simple Login Manager, but then our renamed binary _could_ be masked from users and referred to as `slim` still.

Alternatively, we might be permitted by the Fedora Packaging Committee to use `Conflicts: slim` in our package definition, so that users cannot have both Simple Login Manager and Selection on Linked Mutations installed at the same time. However, since they do not provide similar capabilities I think they will not permit this, as the conflict is circumstantial and not fundamental.

This issue will be referenced in my self-introduction to the devel mailing list of the Fedora Project, and hopefully we can come to a nice solution that works for both those users of Simple Login Manager and the SLiM community (us!), and have SLiMgui accepted into the official repositories soon.

Additionally, I have made significant progress on the Debian & Ubuntu package, which I will also be submitting to official repositories. The package will also be named SLiMgui there, but the decision we come to here regarding the command-line binary will influence how I approach the Debian maintainers.

To summarize, as I know this issue is long and entirely prose, there are three things we can do:

- Rename the `slim` binary to `SLiMcli`. This affects end-users and is inadvisable.
- Request permission to use a `Conflicts: slim` entry in our package definition, so that SLiM and Simple Login Manager cannot be installed on the same system.
- Split SLiMgui into two packages, one containing only SLiMgui, and the other, "SLiMcli" containing the `slim` binary and the `eidos` binary, kicking the conflict down the road.

Regards,

Bryce. 🐙

bryce-carson:
# Summary
- This merge request (MR) updates the RPM and [Appstream](https://www.freedesktop.org/software/appstream/docs/) side of things.
  - The RPM Spec file didn't have a new changelog entry when we released v4.0. That needs to be added.
  - The Appstream metadata that will be consumed by the Flatpak (mentioned in another item of this summary and MR) will display the screenshot uploaded to my fork of SLiM.
- This MR informs you of my intent to release a Flatpak for Debian and Ubuntu users, and addresses the concern(s) you raised in our personal communications.

# Personal communications
Ben, I'm responding to your email in this merge request because my digital life is a bit spread around at the moment. It seems simpler to respond and propose a solution in one place, rather than multiple.

## Response

> Hi Bryce.  ... I don't know the pros and cons of all these different [build, packaging, release] approaches, but the "Flatpak" approach certainly sounds better for Debian/Ubuntu, since it sounds much simpler.  Is there any big drawback?

Hey, Ben.

The pros and cons of different build-types on Copr are limited to the packager, and some of these depend on whether the package being built by Copr has any RPM-related files and metadata in the upstream repository (like SLiM has for a while).

The approach I was using before did not leverage the Git-managed RPM spec file that I pushed to the repository, while the new one does! :) The new approach is a SCM (source-control management?) based build, rather than one where I upload a new version of the spec file to Copr manually. The new approach simply builds "more directly" from Git.

This means that in the future, a release includes:
 1. Update the SPEC file to change the `Version` field, and drop the `Release` field back to `1` for this software version.
 2. On Copr, create a new build that uses the tag, release, or Git hash to point Copr to use that specific version of the sources, and use the HTTPS clone url for the sources.
 
Tangentially related to this new SCM-approach, I'm closing #174 as I don't think it's needed. It would be redundant, and since nothing is expected to change in Qt5, or with the C++ sources of SLiMgui itself, it seems like it would only be headache.

> What's the reason you didn't choose this distribution method before?  Anyhow, as usual I am happy to trust your judgement; if this seems to you like the right path forward (and it sounds like it does), then let's do it!  :->  Let me know when it's time to revise the manual, and I'll send you the chapter as a Pages doc again.  Thanks!

We didn't use the Flatpak approach before for two reasons you mentioned in #175.

> [Flatpak] being containerized seems like it has substantial pros and cons. Pros are consistency and predictability of the build environment; presumably there would be fewer complaints that the install failed because of this or that toolchain/OpenGL/whatever issue. Cons are (a) complexity, (b) perhaps longer install time?, (c) maybe compatibility issues. Regarding (c), here's my concern. Qt recommends different specific versions of Qt for different platforms – different versions of Ubuntu vs. different versions of Debian vs. whatever. Generally, if a platform provides a pre-installed Qt as many do, it is a version of Qt that is deemed optimal for that platform, I suppose. Overriding that with a one-size-fits-all version of Qt inside the container might not be a good idea, and might cause problems that we would be unaware of until users encountered them on a particular platform. Similarly, I imagine that each platform has a standard "toolchain" – compiler, linker, C++ standard library, etc. – that are tested and optimized for that platform. I imagine the container approach overrides that, and again, I'm not sure that's a good idea.

I'll respond to each of these as best as I understand them now:

### Pros of Flatpak
- Consistency and build predictability: indeed!
- Resolve Qt runtime/build-time verison mismatch (if and only if a user elects to use the Flatpak (we retain the RPMs and other options for each platform))
- Ubuntu and Debian have a single, reliable build and release platform that won't trail behind the other releases and will run reliably

### Cons of Flatpak
- A second Flatpak must be released for the CLI `slim`, if at all (the real challenge is releasing SLiMgui in a compiled form for so many platforms; building `slim` from sources on any platform _is_ simple, and I advise it for end users over a second Flatpak)
- Qt recommendations: if Qt recommends different versions of Qt for different platforms, that's based on Qt and that platforms discussions and testing. The way Flatpak works is it bundles a runtime and a Sdk. Applications are build with an Sdk in a container or `chroot` against a specific runtime/platform. Specifically, the runtime that SLiMgui is built against now is Ubuntu 21.08 LTS, so Qt5.15 being the recommended (and last) version of Qt for that platform is satisfied.

At least for Debian and Ubuntu, this is not a drawback. Debian has slightly older packages than Ubuntu, in most cases, and using this allows us to use the same version of Qt and the same runtime libraries and dependencies on each version of Debian and Ubuntu. That makes things dead simple, and we don't need to worry about packaging issues.

I won't recommend Flatpak for _everyone_, but it _will work_ for everyone.

# Flatpak
I'm creating a repository called `SLiM-flatpak`. I need to create a GitHub pages website for that repository so that a file `org.messerlab.slimgui.flatpakref` can be pointed to with a standard URI rather than the "blobby" "raw.githubusercontent.com/" domain URI. That's what I've read must be done when you want to use a GitHub repository as a Flatpak repository.

Eventually, I'll submit another MR to update the README.md on the main MesserLab/SLiM repository to include a button to one-click install the reference to the repository and SLimgui on whatever platform supports Flatpak (that's all of them).

This resolves #175 and also #288. It also resolves https://github.com/MesserLab/SLiM-Extras/issues/22.

The reason this resolves #175 is not because it includes SLiMgui in the official Fedora repos, but because I'm retiring that pursuit for now, and we can always reopen it in the future.


bhaller:
Hi Bryce.  Thanks for your exceedingly clear communication, as always!  I think I understand.  Funny, I don't remember writing that stuff in #175 at all, but that's the problem of course – I kinda get my head wrapped around this stuff at least a little bit, and then a year or more passes in which I don't think about it, and I forget what little I managed to learn.  :->  Anyhow, it all sounds good I think/hope.  I'm a little confused/concerned regarding the Qt version issue; I'm not sure that Qt 5.15 will work on all platforms that people will want to install on, I think some older platforms are probably back on something from 5.9 to 5.14.  But in practice, my experience has been that Qt is less version-sensitive than their documentation suggests, so it might be fine; let's see how it goes.  SLiMgui mostly relies on stable features in Qt that were probably not changing very much in those versions; it is not on the cutting edge of Qt APIs.

So, I take it this is ready to merge, so I'm merging it!  :->

PR #391, opened 2023-09-17
note explaining current status of slim_id in population table

### LINKED ISSUE #390: remove slim_id == population row requirement on reading in tree sequences?
It's not too hard for someone to make a tree sequence not readable by SLiM because of a probably overly-strict check we have on population tables: suppose the SLiM model has only `p1`; they write it out, do something to it, *simplify it*, and try to read it back in. The population table will have had an (empty) 0th row, which would be removed by simplify, so in the resulting tree sequence the 0th row has a SLiM population with `slim_id=1` in the metadata; this triggers [this check](https://github.com/MesserLab/SLiM/blob/931a33f69c7ef9cb87317ae7641151cb6c0643c4/core/species.cpp#L7183).

The right thing to do for users is probably to *not* simplify; or if they do then pass `filter_populations=False`. But, we really probably don't need that check, so we might as well remove it. To do so we just need a few tests, maybe in pyslim, that do weird things to the population table and check nothing bad happens - for instance, make a bunch of populations all with the same `slim_id`; etcetera.

(This showed up in #365.)

petrelharp:
I had a go at removing the requirement that metadata and population row match, but discovered its tentacles reaching out further than I anticipated. Here's the result.

Closes #390 .

### DIFF
--- a/core/species.cpp
+++ b/core/species.cpp
@@ -7177,8 +7177,9 @@ void Species::__RemapSubpopulationIDs(SUBPOP_REMAP_HASH &p_subpop_map, int p_fil
 					// and we will remap it and fix up its metadata
 					slim_objectid_t slim_id = subpop_metadata["slim_id"].get<slim_objectid_t>();
 					
-					// enforce the slim_id == index invariant here; it is not clear that we really need this
-					// invariant, but I'm going to enforce it for now so I don't have to think about it
+					// enforce the slim_id == index invariant here; removing this invariant would be
+					// possible but would require a bunch of bookeeping and checks; see treerec/implementation.md
+                    // for more discussion of this
 					if (slim_id != subpop_id)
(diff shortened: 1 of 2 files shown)


bhaller:
OK.  I'm not surprised to hear that the tentacles have extended through the code.  :-<  But the way it is is not so bad really; it's pretty rare for people to hit a problem with it, at this point, and generally not hard for that problem to be fixed once diagnosed.  Thanks for looking into it.

bhaller:
@petrelharp is this waiting for anything else before being merged?

petrelharp:
Nope - you should read to see if this makes sense.

bhaller:
OK, this looks good to me.  Branch `smooth` has been merged, and this has no conflicts, so I'm merging this.  Onward and upward.

PR #397, opened 2023-10-06
fix sampleNearbyPoint() and sampleImprovedNearbyPoint()

### LINKED ISSUE #396: possible issue in `sampleNearbyPoint`?
In investigating why `sampleNearbyPoint` seems to be going so slooow in some cases, I noticed the following thing that looks wrong to me? Observe that in `sampleNearbyPoint`, the [reference point is scaled by the bounds](https://github.com/MesserLab/SLiM/blob/master/core/spatial_map.cpp#L2322) to lie between 0 and 1. However, to this point [is added a displacement](https://github.com/MesserLab/SLiM/blob/master/core/spatial_map.cpp#L2334) that if I'm reading correctly, is [*not* correspondingly scaled](https://github.com/MesserLab/SLiM/blob/master/core/spatial_kernel.cpp#L359).  (I am probably missing something, though.)

bhaller:
Fix #396.

@petrelharp please review this very carefully; unit testing of this code is difficult, so bugs need to be caught by eye, at least for now.  :-O  Please check that the fixes are parallel and complete for 1D/2D/3D, between both sampleNearbyPoint() and sampleImprovedNearbyPoint().  They look right to me, but I worked quickly because I'm about to be out of time for a while.  Thanks!


petrelharp:
Shouldn't [this bit](https://github.com/MesserLab/SLiM/pull/397/files#diff-b8bc526c0dd677af49f50ae4134662153e6e72482c801886b4f2d0a4ef3dd71fR2114) be
instead of
```
				while (displaced_point[0] < 0.0)
					displaced_point[0] += bounds_a1_;
				while (displaced_point[0] > bounds_a1_)
					displaced_point[0] -= bounds_a1_;
```
how about
```
				while (displaced_point[0] < bounds_a0_)
					displaced_point[0] += bounds_a1_ - bounds_a0_;
				while (displaced_point[0] > bounds_a1_)
					displaced_point[0] -= bounds_a1_ - bounds_a0_;
```
(and in other similar places further down)?

bhaller:
> Shouldn't...

Nope, no need.  When periodic boundaries are enabled, the lower bound of the spatial bounds must be 0.0, and that is enforced throughout the API.  It doesn't really make sense to be periodic out in the middle of the coordinate space, or that's my opinion anyway.  :->  So we can assume 0.0 at those points in the code.

petrelharp:
Do we note in the documentation that `sampleNearbyPoint` will be very inefficient if the maximum value of the map (anywhere) is much larger than the typical value where individuals are? (I think not, but we should?)  Maybe like:
```
The algorithm works by rejection sampling, and so will be very inefficient if the maximum value of the map (anywhere, across the entire map) is much larger than the typical value of the map where individuals are.
```

bhaller:
Hmm.  Seems like a reasonable thing to note; and we could point the user towards sampleImprovedNearbyPoint() as an alternative that does not suffer from this problem, right?  I'd say especially given that we can recommend that alternative, it's worth adding this note...?

core/spatial_map.cpp
```diff
@@ -2346,28 +2358,27 @@ EidosValue_SP SpatialMap::ExecuteMethod_sampleNearbyPoint(EidosGlobalStringID p_
...
+			*(result_ptr++) = displaced_point[0];
 		}
 	}
 	else if (spatiality_ == 2)
 	{
 		// FIXME: TO BE PARALLELIZED
 		for (size_t point_index = 0; point_index < point_count; point_index++)
 		{
-			// scale point coordinates to [0, 1]
 			double point_a = *(point_buf_ptr)++;
```
petrelharp, line 2380: in `sampleImprovedPoint` this was like `double point[X]`, not `point_a`, `point_b`, etcetera. (I don't see anything wrong, just differences in what could be parallel style of code?)

petrelharp (approved):
I don't see anything wrong!

core/spatial_map.cpp
bhaller: Yes.  There was a reason for that, in the old code, but that reason has now evaporated, so I'll shift it to be parallel.

bhaller:
OK, I'll clean up the loose ends you noted, and then merge it in.  Thanks for the quick review!

PR #406, opened 2023-11-08
tskit update to C_1.1.2

petrelharp:
Perhaps this will fix https://github.com/tskit-dev/tskit/issues/2868?


petrelharp:
This seems to work for me - passes pyslim tests and everything. And results in less lost memory over at https://github.com/tskit-dev/tskit/issues/2868.

bhaller:
Merged it – whether it fixes the problem or not, it makes sense to update.

PR #413, opened 2023-12-05
Use CMake to perform installation, and RPM macros

bryce-carson:
Hi Ben,

This PR updates the RPM spec file for the 4.1 release, adopts recommended RPM macros rather than hard-coding things (though that did appear robust in the past), and makes CMake capable of installing SLiMgui, slim, and eidos with desktop integration (which simplifies the role of RPM and makes the spec file more maintainable).

It didn't cross my mind until now, but the following should work beautifully and provide desktop integration on Linux platforms when building from source and should also be integrated into the manual's installation instructions:

```sh
cmake -D BUILD_SLIMGUI=ON
make
make install # This line, on Linux, will now *also* install Desktop integration files.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,9 +348,9 @@ else()
         target_link_libraries(${TARGET_NAME} PUBLIC Qt5::Widgets Qt5::Core Qt5::Gui OpenGL::GL gsl tables eidos_zlib gnu )
     else()
 	    target_link_libraries( ${TARGET_NAME} PUBLIC Qt5::Widgets Qt5::Core Qt5::Gui OpenGL::GL gsl tables eidos_zlib )
-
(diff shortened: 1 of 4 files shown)


bhaller:
Looks good to me, thanks!  I haven't looked at the manual changes you sent yet; I suppose the bit you mention above in your comment has been worked in with your edits.

bryce-carson:
> Looks good to me, thanks! I haven't looked at the manual changes you sent yet; I suppose the bit you mention above in your comment has been worked in with your edits.

Hi Ben,

It hasn't, actually. I sent my edits to you already, before I thought of it.

I will make the change now and re-send the edition, since that will probably be the easiest for both of us (it's easier for me than explaining where to make the edition, and easier for you too, I hope).

Issue #419, opened 2023-12-11
genome id mismatch error with SLiM 4.1


ShyamieG:
EDIT: It turns out this error was caused by my script, which re-labels nodes within the tree sequence to eliminate duplicate slim ids. In SLiM 4.1, a mismatch between node slim id and individual pedigree id throws this error.

Hi again,

I'm running into an issue with at a previously error-free step after updating to SLiM 4.1 (specifically, the version that was just patched to fix [this issue](https://github.com/MesserLab/SLiM/issues/418))

A couple of my steps (i.e. 2 out of >400) now throw this error:

```
ERROR (Species::__TabulateSubpopulationsFromTreeSequence): genome id mismatch; this file cannot be read.
```

I've included the [input files](https://github.com/MesserLab/SLiM/files/13641041/genome_id_issue.zip) and code to reproduce the error:

```
slim -l 0 -seed 15567 -d "IN='H4152'" -d "infile='inf-270_V5490_to_H4152_on_day_9_simplified_annotated.ts'" -d "inf_idx='283'" -d "out=c('V1815')" -d "sampling_events=c(42)" -d "outpath='.'" -d "mu='3.00E-10'" -d "positions_file='PvP01_genome_positions'" -d "N_transmitted='20'" -d "N_reactivated='1'" -d "N_sampled='100'" -d "t_liver_stage='7'" -d "N_liver_final='2.00E+04'" -d "t_to_99_blood_max='7'" -d "t_RBC_period='2'" -d "N_blood_max='1.00E+05'" -d "N_RBC_clones='16'" -d "t_gameto_lifespan='5'" -d "p_RBC_to_gameto='0.01'" -d "t_start_clearance='40'" -d "t_inf_half_life='50'" hhar.slim
```

In case it is helpful, here is another example that runs without an error:

```
[...]

Issue #416, opened 2023-12-06
error adding individuals to table


ShyamieG:
I'm running into a strange issue running this script (input files are [attached here](https://github.com/MesserLab/SLiM/files/13589937/slim_issue.zip)):

```
slim -l 0 -seed 14081 -d "IN='H5834'" -d "infile='inf-1852_V8157_to_H5834_on_day_23_simplified_annotated.ts'" -d "inf_idx='1893'" -d "out=c('V1945','V105','sample')" -d "sampling_events=c(21,28,61)" -d "outpath='.'" -d "mu='3.00E-10'" -d "positions_file='PvP01_genome_positions'" -d "N_transmitted='20'" -d "N_reactivated='1'" -d "N_sampled='100'" -d "t_liver_stage='7'" -d "N_liver_final='2.00E+04'" -d "t_to_99_blood_max='7'" -d "t_RBC_period='2'" -d "N_blood_max='1.00E+05'" -d "N_RBC_clones='16'" -d "t_gameto_lifespan='5'" -d "p_RBC_to_gameto='0.01'" -d "t_start_clearance='40'" -d "t_inf_half_life='50'" hhar.slim
```

When running this on my institution's cluster, I get a tskit out of memory error:

```
Error:tsk_individual_table_add_row: Out of memory. (TSK_ERR_NO_MEMORY)
```

However, when I moved the input files to my local machine to better troubleshoot, I saw a different error:

```
Assertion failed: (p_tables->nodes.individual[ind->genome1_->tsk_node_id_] == (tsk_id_t)tsk_individual), function AddIndividualsToTable, file /Users/bhaller/Desktop/SLiM/core/species.cpp, line 4970.
zsh: abort      slim -l 0 -seed 14081 -d "IN='H5834'" -d  -d "inf_idx='1893'" -d  -d  -d  -d
```

I sought help from my institution's IT department to troubleshoot the error on the cluster, but wasn't able to solve it (increasing the memory request did not help). Have you encountered this kind of issue before?

Here is an example using the same underlying slim script (but different input file) without issue:

```
[...]

PR #420, opened 2023-12-13
add node pedigree ID check

bhaller:
Related to #416 and #419, this PR adds a check, on tree sequence load, for nodes with pedigree IDs that (once divided by two to get their corresponding individual pedigree ID) are higher than any individual pedigree ID (meaning that the next pedigree ID used needs to be higher).  At present this PR checks for this condition, emits a warning if it sees it, and fixes the problem by setting the next pedigree ID to be sufficiently high.  We might want to change this to be an error condition instead, if we can't think of any case when it would legitimately occur.

### DIFF
--- a/core/species.cpp
+++ b/core/species.cpp
@@ -7684,6 +7684,7 @@ void Species::__CreateSubpopulationsFromTabulation(std::unordered_map<slim_objec
 {
 	// We will keep track of all pedigree IDs used, and check at the end that they do not collide; faster than checking as we go
 	// This could be done with a hash table, but I imagine that would be slower until the number of individuals becomes very large
+	// Also, I'm a bit nervous about putting a large number of consecutive integers into a hash table, re: edge-case performance
 	std::vector<slim_pedigreeid_t> pedigree_id_check;
 	
 	gSLiM_next_pedigree_id = 0;
@@ -7809,12 +7810,12 @@ void Species::__CreateSubpopulationsFromTabulation(std::unordered_map<slim_objec
 		}
 	}
 	
-	// Check for pedigree ID collisions by sorting and looking for duplicates
+	// Check for individual pedigree ID collisions by sorting and looking for duplicates
 	std::sort(pedigree_id_check.begin(), pedigree_id_check.end());
 	const auto duplicate = std::adjacent_find(pedigree_id_check.begin(), pedigree_id_check.end());
 	
 	if (duplicate != pedigree_id_check.end())
-		EIDOS_TERMINATION << "ERROR (Species::__CreateSubpopulationsFromTabulation): the pedigree ID value " << *duplicate << " was used more than once; pedigree IDs must be unique." << EidosTerminate();
+		EIDOS_TERMINATION << "ERROR (Species::__CreateSubpopulationsFromTabulation): the individual pedigree ID value " << *duplicate << " was used more than once; individual pedigree IDs must be unique." << EidosTerminate();
 }
 
 void Species::__ConfigureSubpopulationsFromTables(EidosInterpreter *p_interpreter)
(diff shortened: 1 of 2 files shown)


bhaller:
OK @petrelharp, it is very easy to make this happen with a pure SLiM model:

```
initialize() {
	initializeSLiMModelType("nonWF");
	initializeTreeSeq();
	initializeMutationType("m1", 0.5, "f", 0.0);
	initializeGenomicElement(g1, 0, 99999);
	initializeMutationRate(1e-4);
	initializeRecombinationRate(1e-4);
}
1 late() {
	sim.addSubpop("p1", 1);
}
reproduction() {
	subpop.addCloned(p1.individuals);
}
2 late() {
	// The goal is to end up with genomes in the tree sequence that are more
	// recent (with a larger pedigree ID) than any extant individual.  So we
	// just kill the offspring; its genomes are still in the tree sequence
	// since we don't simplify on save, and they are younger than the extant
	// individual.
	offspring = p1.subsetIndividuals(maxAge=0);
	sim.killIndividuals(offspring);
	
	sim.treeSeqOutput("~/Desktop/test_ts.trees", simplify=F);
}
```

You don't even need to do `rememberIndividuals()`.  The only thing needed is to kill the offspring generation, and save with `simplify=F`.  You then load it back in like so:

```
initialize() {
	initializeSLiMModelType("nonWF");
	initializeTreeSeq();
	initializeMutationType("m1", 0.5, "f", 0.0);
	initializeGenomicElement(g1, 0, 99999);
	initializeMutationRate(1e-4);
	initializeRecombinationRate(1e-4);
}
2 late() {
	sim.readFromPopulationFile("~/Desktop/test_ts.trees");
}
```

With the code in this PR that does the ID check, you then get:

`#WARNING (Species::__CheckNodePedigreeIDs): in reading the tree sequence, a node was encountered with a genome pedigree ID that was (after division by 2) greater than or equal to the next pedigree ID; this is not expected to happen.`

So it's not an error condition.  At worst it's a warning; I'm not even sure we ought to do that, really.  The argument for giving the warning would be that this is expected to be rare, and so it is more likely to be a problem than to be legitimate.  I guess using simplify=F is pretty rare, so I guess I'll keep it as a warning?  What do you think?  I'm tempted to just say "There is nothing wrong, so it shouldn't warn."  Anyhow, with the code in this PR we now fix things so the next pedigree ID used by the simulation is correct, so if this is the only "problem" with the file (which isn't truly a problem at all), things will now work correctly in this case.

Tomorrow I plan to add code to this PR to check for duplicate genome pedigree IDs, so that we cover that base as well.  That, of course, would be a stone cold error.

bhaller:
Added the check for duplicate genome pedigree IDs.

Changed the warning message for the genome-id-unexpectedly-large case to read: "`#WARNING (Species::__CheckNodePedigreeIDs): in reading the tree sequence, a node was encountered with a genome pedigree ID that was (after division by 2) greater than the largest individual pedigree ID in the tree sequence.  This is not necessarily an error, but it is highly unusual, and could indicate that the tree sequence file is corrupted.  It typically only happens if an unsimplified tree sequence is being loaded, and even then only under certain conditions.`"

This PR is done, except for decided whether we want that to be a warning at all.  @petrelharp please review, thanks.

petrelharp:
A warning seems fine. How about:
```
in reading the tree sequence, a node was encountered with a genome pedigree ID that was (after division by 2) greater than the largest individual pedigree ID in the tree sequence.  This is not necessarily an error, but it is highly unusual, and could indicate that the tree sequence file is corrupted.  It may happen due to external manipulations of a tree sequence, or  perhaps if an unsimplified tree sequence produced by SLiM is being loaded.
```

petrelharp:
>  I guess you'll need to construct those pathological tree sequence files in Python, they should be impossible to make in SLiM. 

How do you want these? (python code or .trees file?)

bhaller:
> > I guess you'll need to construct those pathological tree sequence files in Python, they should be impossible to make in SLiM.
> 
> How do you want these? (python code or .trees file?)

I was thinking they'd go into the tests that you maintain, not into my unit tests, no?  My tests don't have any way to incorporate either Python code or a .trees file.

core/species.cpp
```diff
@@ -8398,6 +8399,64 @@ void Species::__AddMutationsFromTreeSequenceToGenomes(std::unordered_map<slim_mu
...
+	for (tsk_size_t j = 0; (size_t)j < node_count; j++)
+	{
+		tsk_size_t offset1 = node_table.metadata_offset[j];
+		tsk_size_t offset2 = node_table.metadata_offset[j + 1];
+		tsk_size_t length = (offset2 - offset1);
+		
+		if (length)		// allow nodes with no metadata, which might be carryover non-SLiM nodes
+		{
+			// but if a node has metadata, it must be SLiM metadata, I think; our schema guarantees that
+			if (length != sizeof(GenomeMetadataRec))
```
petrelharp, line 8425: I think this is inserting a new requirement on the loaded tree sequence that might well come back to bite us - previously, we didn't care what the metadata was for ancestral nodes.  I think you may be right about the metadata schema, but if we don't need to be more picky about metadata than we were before, let's not start?
```suggestion
		if (length != sizeof(GenomeMetadataRec))
```

petrelharp: I realize there is now the possibility that there is metadata of the right length but the wrong structure, that errors, but this seems pretty unlikely.

bhaller: So, if the metadata length == sizeof(GenomeMetadataRec) then assume it is our metadata and go into it, if it is any other length then just ignore it?  I can do that, sure, although eww.

bhaller: But yeah, I see your point, and length is the only criterion we really have for checking whether metadata is ours or not, so there we are.

petrelharp (approved):
One comment, otherwise good!

Issue #438, opened 2024-04-09
provide `make test` functionality to run tests after building `slim` and `eidos`


bhaller:
@bryce-carson wrote in #437 (the `*** buffer overflow detected ***` fallout from the `-D_FORTIFY_SOURCE=3` change in some Linux systems):

> @bhaller, if you can write a new section in the Makefile to run the tests, so that make test can be called, that'd be useful for the packaging process to detect issues like these in the future (it's not needed for the patched RPM; that's already live). I tried to locate the binaries and call the tests myself, but I couldn't find them where I expected them to be. It's easier to hand the work off to CMake since it magically knows where the binaries are.

This issue represents that request.  Presumably a fairly simple CMake change, but I'm not very fluent with CMake so who knows.  :->  If somebody else reading this knows what to do (tagging @grahamgower as well), feel free to submit a PR.  :->

bhaller:
OK, with #449 now merged, I think this issue is now fixable...?  Let me know if there's any obstacle remaining.  Thanks!

bhaller:
I guess this issue is completed and should be closed.  @bryce-carson if you do intended to use this `make test` functionality to make the RPM builds more robust, as discussed elsewhere, please feel free to make a new issue, or just have it on your own personal to-do list, as you wish.  :->  Thanks all!

Issue #437, opened 2024-04-02
"buffer overflow detected" when trying to install SLiM on Linux


alfroids:
I've been having trouble when trying to install SLiM from the AUR using paru in Arch. The package seems to be downloaded and built correctly, but when installation tests start (specifically `-testSLiM`), I get a `*** buffer overflow detected ***` and the installation is aborted. The installation log ends like this:

```
[100%] Built target SLiMgui
==> Starting check()...
Running 'slim -testEidos'...

SUCCESS count: 6953
Running 'slim -testSLiM'...
*** buffer overflow detected ***: terminated
/home/alfroids/.cache/paru/clone/slim-simulator/PKGBUILD: line 42:  8003 Aborted                 (core dumped) ./slim -testSLiM
SLiM tests failed
==> ERROR: A failure occurred in check().
    Aborting...
error: failed to build 'slim-simulator-4.2.r0.g6d00ad37-1':
error: packages failed to build: slim-simulator-4.2.r0.g6d00ad37-1
```

bhaller:
Hi @alfroids.  I'm going to ping @grahamgower since he maintains the Arch installer, and is an Arch user himself.  I don't know what might be going on here.  The text "buffer overflow detected" does not occur in the SLiM code base, I believe, so I don't think this is a SLiM error.  So I'm not sure whose error it is!  Somebody's buffer is overflowing and they're complaining about it, but I'm not sure whether it has anything to do with SLiM or not.

A couple of questions.  (1) Have you tried doing this multiple times, and it always fails the same way each time?  (2) Does the failure of the installer test mean that SLiM gets removed, or does it remain installed on your machine?  If it remains installed on your machine, can you run `slim -testSLiM` yourself, and does it succeed?  (3) Can you look at `/home/alfroids/.cache/paru/clone/slim-simulator/PKGBUILD` and see what is at line 42?

It would be nice to track this down, but I guess nobody else on Arch (including @grahamgower himself) has hit this error, so it might be something specific to your machine.  Anything unusual about it?  Is it a real Arch box, or is it a virtual machine running under something-or-other?  Does it have a full, standard Arch install?  What version, etc.?

alfroids:
(1) Yes, I tried multiple times today and yesterday and always got the same result.
(2) The files and scripts are still there in the same `paru/clone/slim-simulator` directory and I can run them directly. SLiMgui runs and the default recipe works fine, but when I run `./slim -testSLiM` I get the same message as before, just shorter.
[...]

PR #449, opened 2024-07-03
Add make test feat

currocam:
I'm no expert, but I think this would solve #438 

I tested it by doing:

1. Generate build system
```bash
cd SLiM
mkdir build
cd build
cmake ..
```
2. Run test before building
```bash
make test
```
3. Clean files, build and then test
```bash
make clean
make
make test
``` 
4. Edit some of the test files and repeat 1-3 to assert the tests actually fail when they should

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -427,14 +427,7 @@ endif()
 
 # add testing so it can be called using make test
 enable_testing()
-# we add test_build to generate the slim binary if make build has not been called first
-add_test(test_build
-  "${CMAKE_COMMAND}"
-  --build "${CMAKE_BINARY_DIR}"
-  --config "$<CONFIG>"
-  --target ${TARGET_NAME_SLIM}
-)
-set_tests_properties(test_build PROPERTIES FIXTURES_SETUP test_fixture)
+
 # test SLiM
 add_test(
   NAME testSLiM
@@ -445,6 +438,3 @@ add_test(
   NAME testEidos
   COMMAND ${TARGET_NAME_SLIM} -testEidos
 )
-set_tests_properties(testSLiM PROPERTIES FIXTURES_REQUIRED test_fixture)
-set_tests_properties(testEidos PROPERTIES FIXTURES_REQUIRED test_fixture)
-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,42 +234,42 @@ if(BUILD_LTO)
 endif()
 
 # GSL - adding /usr/local/include so all targets that use GSL_INCLUDES get omp.h
-set(TARGET_NAME gsl)
+set(TARGET_NAME_GSL gsl)
 file(GLOB_RECURSE GSL_SOURCES ${PROJECT_SOURCE_DIR}/gsl/*.c ${PROJECT_SOURCE_DIR}/gsl/*/*.c)
 set(GSL_INCLUDES ${PROJECT_SOURCE_DIR}/gsl ${PROJECT_SOURCE_DIR}/gsl/specfunc ${PROJECT_SOURCE_DIR}/gsl/blas ${PROJECT_SOURCE_DIR}/gsl/rng ${PROJECT_SOURCE_DIR}/gsl/cdf ${PROJECT_SOURCE_DIR}/gsl/vector ${PROJECT_SOURCE_DIR}/gsl/err ${PROJECT_SOURCE_DIR}/gsl/sys ${PROJECT_SOURCE_DIR}/gsl/randist ${PROJECT_SOURCE_DIR}/gsl/matrix ${PROJECT_SOURCE_DIR}/gsl/cblas ${PROJECT_SOURCE_DIR}/gsl/complex ${PROJECT_SOURCE_DIR}/gsl/block ${PROJECT_SOURCE_DIR}/gsl/interpolation ${PROJECT_SOURCE_DIR}/gsl/linalg /usr/local/include)
-add_library(${TARGET_NAME} STATIC ${GSL_SOURCES})
-target_include_directories(${TARGET_NAME} PUBLIC ${GSL_INCLUDES})
+add_library(${TARGET_NAME_GSL} STATIC ${GSL_SOURCES})
+target_include_directories(${TARGET_NAME_GSL} PUBLIC ${GSL_INCLUDES})
 
 # ZLIB
-set(TARGET_NAME eidos_zlib)
+set(TARGET_NAME_ZLIB eidos_zlib)
 file(GLOB_RECURSE ZLIB_SOURCES ${PROJECT_SOURCE_DIR}/eidos_zlib/*.c)
 set(ZLIB_INCLUDES ${PROJECT_SOURCE_DIR}/eidos_zlib)
-add_library(${TARGET_NAME} STATIC ${ZLIB_SOURCES})
-target_include_directories(${TARGET_NAME} PUBLIC)
+add_library(${TARGET_NAME_ZLIB} STATIC ${ZLIB_SOURCES})
+target_include_directories(${TARGET_NAME_ZLIB} PUBLIC)
 
 # KASTORE
-set(TARGET_NAME kastore)
+set(TARGET_NAME_KASTORE kastore)
 file(GLOB_RECURSE KASTORE_SOURCES ${PROJECT_SOURCE_DIR}/treerec/tskit/kastore/*.c)
 set(KASTORE_INCLUDES ${PROJECT_SOURCE_DIR}/treerec/tskit/kastore)
-add_library(${TARGET_NAME} STATIC ${KASTORE_SOURCES})
-target_include_directories(${TARGET_NAME} PUBLIC)
+add_library(${TARGET_NAME_KASTORE} STATIC ${KASTORE_SOURCES})
+target_include_directories(${TARGET_NAME_KASTORE} PUBLIC)
 
 # TSKIT
-set(TARGET_NAME tables)
+set(TARGET_NAME_TSKIT tables)
 file(GLOB_RECURSE TABLE_SOURCES ${PROJECT_SOURCE_DIR}/treerec/tskit/*.c)
 set(TSKIT_INCLUDES ${PROJECT_SOURCE_DIR}/treerec)
-add_library(${TARGET_NAME} STATIC ${TABLE_SOURCES})
-target_include_directories(${TARGET_NAME} PRIVATE ${GSL_INCLUDES} ${KASTORE_INCLUDES})
-target_include_directories(${TARGET_NAME} PUBLIC ${KASTORE_INCLUDES} ${TSKIT_INCLUDES})
+add_library(${TARGET_NAME_TSKIT} STATIC ${TABLE_SOURCES})
+target_include_directories(${TARGET_NAME_TSKIT} PRIVATE ${GSL_INCLUDES} ${KASTORE_INCLUDES})
+target_include_directories(${TARGET_NAME_TSKIT} PUBLIC ${KASTORE_INCLUDES} ${TSKIT_INCLUDES})
 if(WIN32)
-    target_link_libraries(${TARGET_NAME} PUBLIC gnu)
+    target_link_libraries(${TARGET_NAME_TSKIT} PUBLIC gnu)
 endif()
 
 # SLIM
 if(PARALLEL)
-	set(TARGET_NAME slim_multi)
+	set(TARGET_NAME_SLIM slim_multi)
 else()
-	set(TARGET_NAME slim)
+	set(TARGET_NAME_SLIM slim)
 endif(PARALLEL)
 file(GLOB_RECURSE SLIM_SOURCES ${PROJECT_SOURCE_DIR}/core/*.cpp ${PROJECT_SOURCE_DIR}/eidos/*.cpp)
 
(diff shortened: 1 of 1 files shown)


bhaller:
@bryce-carson thoughts?

bhaller:
Thanks for the contribution @currocam!

bryce-carson:
@bhaller, I am not familiar with writing tests in CMake either, so this looks alright to me.

I'd like to see GitHub workflows updated to take advantage of the CMake tests, however, rather than calling the tests manually. Ideally, `make test` should be robust enough to detect any errors on all systems.

I'm not familiar with GitHub workflows either, however; I've played with them before, but it was a bit of a struggle for me then. I won't say that we should delay accepting the PR until we integrate `make test` into the workflows, but it's something we should open an issue for to track progress on at least.

currocam:
Hi, 
I agree that it would be nice to use make test in the CI also, as it would be less "surprising" for people looking at the code from outside.

I hesitated to included in the PR because using slim -testSlim and slim -testEidos seems idiomatic enough to me and I'm not sure it would be worthy to add an abstraction just to avoid those two commands

bhaller:
OK.  I'll deal with this once I'm not on vacation.  I think I know how to fix the CI to use it; agreed that that is a minor concern really, but it'd be nice forthe CI to test "make test" itself, so that's a compelling reason to make that change.  I'll circle around to this within a month-ish -- also going to Evolution in Montreal soon, so life is a bit busy right now.  Thanks folks!

bryce-carson:
> [I'm] going to Evolution in Montreal

I'm pretty sure Jon Mee will be there presenting the work I did with him and Sam Yeaman. If you see him there say hello and snag a picture!

bhaller:
Hi folks.  Back from my vacation.  This PR makes sense to me and looks clean, thanks!

The problem is that at least some features of it require a later CMake version than we otherwise require.  For example, support for `FIXTURES_SETUP` and FIXTURES_REQUIRED` was apparently (the CMake doc says, https://cmake.org/cmake/help/v3.30/prop_test/FIXTURES_SETUP.html) only added in CMake 3.7.  For regular (non-SLiMgui) builds, we presently require just CMake 2.8.12.  The goal here, I guess, is to enable a fix for #438.  So, two questions.

One, @currocam, could this "make test" support be made conditional, executed only for CMake version 3.7 and later?  That looks easy (https://stackoverflow.com/a/43697419/2752221), I can do it after merging this PR if you're busy.  Note that `VERSION_GREATER_EQUAL`, which would be the natural thing to use, was itself added in 3.7 and thus cannot be used; it looks like doing an if with `VERSION_LESS` that does nothing, and then doing this test stuff in the else clause of that if (to get `VERSION_GREATER_EQUAL` behavior without using `VERSION_GREATER_EQUAL`) is necessary, sigh.  (Or use `VERSION_GREATER` with 3.6.9 or some such; 3.6.3 was the last 3.6, unlikely there will ever even be a 3.6.4.)

Two, @bryce-carson, will the build system that you want to use this feature in, for #438, actually have a sufficiently recent CMake version present, in all cases?  If not, then this whole approach will not work for your purposes and we need a different approach.

Given this, I might not change the CI workflow to use the new "make test" facility, since I think some of the older versions we test on don't have that recent of a CMake version.  I'm not sure how to check that directly, but I guess I can simply try switching over, and see which CI builds break :->.  But even if we only use this facility in Bryce's stuff for #438, and don't otherwise document/support it (because of the version requirement), that's fine with me; I don't feel that this needs to be user-visible, really.  Nobody has ever asked for it.  Thoughts?

bryce-carson:
Hi @bhaller,

The feature would only be used in the RPM builds, actually.

Here's some references which show, in the first document, how and why to use cmake tests in a spec file, and the second is the CMake documentation which explans how the feature taken advantage of in the first document is used.

*   https://docs.fedoraproject.org/en-US/packaging-guidelines/CMake/
*   [https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html#how-does-cmake-facilitate-testing](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html#how-does-cmake-facilitate-testing)

It really is just a "quality of life" thing for the RPM spec file so that it only reads as

```specfile
%prep
%setup -q

%build
%cmake -DBUILD_SLIMGUI=ON
%cmake_build

%install
%cmake_install

%check
%ctest
```

Rather than

```specfile
%prep
%setup -q

%build
%cmake -DBUILD_SLIMGUI=ON
%cmake_build

%install
%cmake_install
```

With the latter, present code, I don't even run the tests during the RPM build.

bryce-carson:
It seems like there's too much overhead to worry about with this, so I'm unsure if we should bother. The CI builds are stable right now, and I can also call the tests manually in the RPM spec file with as many (or one more) new lines as the following adds.

```specfile
%check
%ctest
```

bhaller:
Hi @bryce-carson,

> The feature would only be used in the RPM builds, actually.
> ...
> It really is just a "quality of life" thing for the RPM spec file...
> With the latter, present code, I don't even run the tests during the RPM build.

Right, I understand this, I think.  Although I'd say it's more than just "quality of life"; #437 occurred because the tests weren't running during the RPM build, right?  And that was a rather unpleasant bug to deal with since we didn't discover it until post-release.  So the goal is to not have that happen again.  :->

My question to you was: what version of CMake is involved in these RPM builds?  Can we rely on it being recent enough (3.7 or later, seems like?) that @currocam's changes to the CMake machinery will run without version errors?  If the answer to that is no, then we need a different approach than the one taken in this PR.

bhaller:
> It seems like there's too much overhead to worry about with this, so I'm unsure if we should bother. The CI builds are stable right now, and I can also call the tests manually in the RPM spec file with as many (or one more) new lines as the following adds.
> 
> ```rpm-spec
> %check
> %ctest
> ```

Not sure I'm understanding this comment.  A PR like this one is needed to make that work, right?  Something needs to bind a "run tests" action to the actual execution of "slim -testEidos" and "slim -testSLiM".  Not sure if we're having communication issues here or what.  Should we zoom?

currocam:
Hi!

Thank you for taking the time to look at this :). I'm having a hard time following the conversation also, but I've a few things I'd like to say:


1. I didn't consider the Cmake version. The FIXTURES_SETUP feature is only used as a "tricky" way to make `make test` work even if you have not run `make` before. I think we can remove it without any issue.  I think test are available in [cmake v2.8.12](https://cmake.org/cmake/help/v2.8.12/cmake.html#section_PropertiesonTests)
2. If the `%ctest` macro executes the test after building the binary, then there is no need of the `FIXTURES_SETUP` to be used, as in the CI one can simply make sure to build the project before running the test (I think)
3. I would gladly make a PR (separate or not) updating the CI, and we can check if it is worthy or not.

bhaller:
> Hi!
> 
> Thank you for taking the time to look at this :). I'm having a hard time following the conversation also, but I've a few things I'd like to say:
> 
> 1. I didn't consider the Cmake version. The FIXTURES_SETUP feature is only used as a "tricky" way to make `make test` work even if you have not run `make` before. I think we can remove it without any issue.  I think test are available in [cmake v2.8.12](https://cmake.org/cmake/help/v2.8.12/cmake.html#section_PropertiesonTests)
> 2. If the `%ctest` macro executes the test after building the binary, then there is no need of the `FIXTURES_SETUP` to be used, as in the CI one can simply make sure to build the project before running the test (I think)
> 3. I would gladly make a PR (separate or not) updating the CI, and we can check if it is worthy or not.

Hi @currocam.  Thanks!  Yes, removing the fixture stuff seems great, if it was not needed.  I just checked the cmake 2.8.12 doc, and it looks like the other stuff you're using was all present that far back, so we should be good to go.  I will merge this now.

bhaller:
And sure, @currocam, go ahead and make a PR that switches the CI over to using this; I appreciate the help.  It ought to work for all targets, since it works with cmake 2.8.12, so there should be no problems with it at all.  Thanks!

bryce-carson:
> what version of CMake is involved in these RPM builds?  Can we rely on it being recent enough (3.7 or later, seems like?) that @currocam's changes to the CMake machinery will run without version errors?

For Arch Linux we can rely on this. Arch has, essentially, the newest software possible for almost every package they ship in their official repository, and most AUR packages are built from the latest upstream source. We can rely on it having a CMake as recent as Fedora Rawhide if not the development version of CMake. Arch does not use the RPM, to be clear.

For Fedora, Red Hat Enterprise, and openSUSE (which all use the RPMs from Copr) we can rely on that.

For Ubuntu we cannot, as our earliest supported version needed the patch I introduced to deal with CMake versions (specifically).

For Debian we cannot, because likewise our oldest supported version has even older package versions than Ubuntu (if I recall correctly).

bryce-carson:
> > It seems like there's too much overhead to worry about with this, so I'm unsure if we should bother. The CI builds are stable right now, and I can also call the tests manually in the RPM spec file with as many (or one more) new lines as the following adds.
> > 
> > ```rpm-spec
> > %check
> > %ctest
> > ```
> 
> Not sure I'm understanding this comment.  A PR like this one is needed to make that work, right?  Something needs to bind a "run tests" action to the actual execution of "slim -testEidos" and "slim -testSLiM".  Not sure if we're having communication issues here or what.  Should we zoom?


That's an RPM macro which just calls ctest, the executable shipped with CMake to run tests registered with CMake in the CMakeLists.txt file.

I can edit the RPM spec to simply run `slim -testSLiM` without relying on this PR at all, so CMake versions won't be any issue.

Arch doesn't use RPM, so @grahamgower just needs to include the lines to test slim and eidos if they aren't already there.

This mostly benefits CI to catch issues if we add a Fedora build, an Arch build, and more Ubuntu and Debian builds.

For packaging, it doesn't really benefit us at the moment (I think).

bhaller:
> > what version of CMake is involved in these RPM builds?  Can we rely on it being recent enough (3.7 or later, seems like?) that @currocam's changes to the CMake machinery will run without version errors?
> 
> For Arch Linux we can rely on this. Arch has, essentially, the newest software possible for almost every package they ship in their official repository, and most AUR packages are built from the latest upstream source. We can rely on it having a CMake as recent as Fedora Rawhide if not the development version of CMake. Arch does not use the RPM, to be clear.
> 
> For Fedora, Red Hat Enterprise, and openSUSE (which all use the RPMs from Copr) we can rely on that.
> 
> For Ubuntu we cannot, as our earliest supported version needed the patch I introduced to deal with CMake versions (specifically).
> 
> For Debian we cannot, because likewise our oldest supported version has even older package versions than Ubuntu (if I recall correctly).

OK; sounds like the CMake approach is not a general solution then.  :-<

bhaller:
> > ...
> 
> That's an RPM macro which just calls ctest, the executable shipped with CMake to run tests registered with CMake in the CMakeLists.txt file.
> 
> I can edit the RPM spec to simply run `slim -testSLiM` without relying on this PR at all, so CMake versions won't be any issue.

Aha, that sounds like a good idea.  I guess we pursued this approach because you wrote on #438: "@bhaller, if you can write a new section in the Makefile to run the tests, so that make test can be called, that'd be useful for the packaging process to detect issues like these in the future (it's not needed for the patched RPM; that's already live). I tried to locate the binaries and call the tests myself, but I couldn't find them where I expected them to be. It's easier to hand the work off to CMake since it magically knows where the binaries are."  But if we can call the tests directly without relying on CMake, that seems clearly better.

> Arch doesn't use RPM, so @grahamgower just needs to include the lines to test slim and eidos if they aren't already there.
> 
> This mostly benefits CI to catch issues if we add a Fedora build, an Arch build, and more Ubuntu and Debian builds.
> 
> For packaging, it doesn't really benefit us at the moment (I think).

Huh.  I guess I misunderstood this also; I thought the reason that #437 was not detected until we shipped was that the tests were not being run during the RPM build, and that #438 was about adding the tests to the RPM build to prevent such a thing from happening to us again.  But that is not the case?  Sorry, I'm really having trouble following this; I guess I totally misunderstood the point of all this, then.  :-<  Well, there is certainly no harm to having merged this PR, even if it now seems to be rather useless; at least it cleaned up the symbol names in the CMake file.  But is there a way for us to actually avoid having #437 type bugs bite us post-release?  To me, that is the actual goal here.

bryce-carson:
I'll need to spend some time reviewing things more intimately next weekend, or during an weekday evening, but #437 initially occurred in an Arch Linux build that @alfroids was using. @jshelly then ran into the issue with a _special_ build of Fedora. It might be that the new default in GCC which sets the fortify\_source level to 3 is present in that container build of Fedora, probably based off of Fedora Rawhide (bleeding edge, nightly builds of Fedora's upstream, untested sources).

I don't have the issue in a stable release of Fedora, and _I don't recall it occurring_ during the COPR build of Fedora Rawhide, but I'll check the logs and what I said in the conversation details when I review this issue more closely later.

bryce-carson:
Actually, @bhaller, I looked at the comment @jshelly made about their container, and they were running Fedora 39. I don't know why they ran into the issue. Perhaps the RPM packages have been broken since Fedora 39 and that was the first report of it that we had! I hope not! The software works on my machine after building from COPR and installing locally, so I don't know what the issue was, other than the forty\_source=3 issue and the tests not being run.

> Indeed. The tests were not run in the RPM package, which I'm really wondering why now. I know I previously had tests run, but I must've removed it when I refactored the spec file.
> 
> I tried adding them back in, but there was some difficulty in finding the binaries. I'll try again next Saturday.

After reading through some of my comments in that thread, I now remember why I wanted `make test`. I couldn't, despite a couple hours interest in it, find the binaries that were built. I ran multiple test runs and did various searches online, but the binaries weren't where I expected them to be and therefore _I couldn't run the tests manually_.

Now that we have `make test` (we do now, right?) then I'll definitely update the packages so that the tests are run on every COPR build.

bhaller:
Hi @bryce-carson.  Sounds like maybe it was more narrow in its scope than we realized, or maybe just nobody reported it.  Well, I wouldn't spend time tracing through that history and trying to figure that out; we fixed that issue and it's done with.  I think the goal now can just be to make sure that tests run during RPM and other installer/package builds, to avoid unexpected failures downstream in future.

Yeah, finding the binaries was the motivation for `make test`, as I understood it; CMake knows where it puts things.  And yes, we now have `make test`, and it is now used in our GitHub Actions CI and works correctly, thanks to @currocam (see PR #455 for the final steps).  And it does not require a later version of CMake than we otherwise require; he cleaned up that issue.  So it should be ready for you to use for COPR.  Thanks!

bryce-carson:
> Hi @bryce-carson.  Sounds like maybe it was more narrow in its scope than we realized, or maybe just nobody reported it.  Well, I wouldn't spend time tracing through that history and trying to figure that out; we fixed that issue and it's done with.  I think the goal now can just be to make sure that tests run during RPM and other installer/package builds, to avoid unexpected failures downstream in future.
> 
> Yeah, finding the binaries was the motivation for `make test`, as I understood it; CMake knows where it puts things.  And yes, we now have `make test`, and it is now used in our GitHub Actions CI and works correctly, thanks to @currocam (see PR #455 for the final steps).  And it does not require a later version of CMake than we otherwise require; he cleaned up that issue.  So it should be ready for you to use for COPR.  Thanks!

Great!

Issue #454, opened 2024-07-15
conda caching does not seem to be working for macos-13


bhaller:
Hi @petrelharp.  You can see that building on macos-13 is not working.  GitHub recently deactivated/removed macos-11; it's time to get macos-13 working (which it never has).  An example of the failure can be seen here:

https://github.com/MesserLab/SLiM/actions/runs/9931045744/job/27430479498

The problem is with conda; you can see the error:

`/Users/runner/.profile: line 22: conda: command not found`

(Just above that there is a "TODO", by you I think, that might also be relevant.)  This error is mystifying to me, though, because earlier in the log are the steps for "Cache conda and dependencies" and "Install conda".  The log says "Cache restored successfully", and it seems to think it succeeded.  So why can't it find conda later on?  And why does this problem only manifest on macos-13, not macos-12?

This kind of thing mystifies me.  Since you put the conda caching stuff in, and perhaps understand it, can you please fix this problem?  Thankyouthankyouthankyou.

bhaller:
Hey, for some reason I can assign issues to you now!  Woot!

bhaller:
If it makes the design/maintenance of this conda stuff simpler, I wonder: is it really worth using this cache mechanism?  A GitHub admin claimed, in a thread I was reading to try to orient myself on this problem, that installing conda takes ~20 seconds (miniconda, in particular, I guess; not sure whether that suffices for our purposes).  That is not huge; maybe if the caching is giving us difficulties it isn't worth it?  But I leave that up to you.  :->

bhaller:
This got fixed by @currocam in #457, closing.

PR #455, opened 2024-07-15
Use make test in CI

currocam:
As we agreed yesterday, here's a minimal PR using `make test` in the CI. I have also modified the Cmake file to also run the Eidos test suite from the Eidos executable (to match previous CI).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -123,9 +123,7 @@ jobs:
       - name: Test (Debug)
         run: |
           cd Debug
-          ./eidos -testEidos
-          ./slim -testEidos
-          ./slim -testSLiM
+          make test
 
       - name: Build (Release)
         run: |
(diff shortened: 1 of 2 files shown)


currocam:
I've done some work git GitHub Actions before, so I might give it a try!

> Perfect, thanks @currocam!
> 
> If you want another project (lol), as you can see the macos-13 CI is failing (and always has, when I have tried it before) because conda is not installed by GitHub Actions any more on macos CI builds, starting with macos-13 (see, e.g., [actions/runner-images#9262](https://github.com/actions/runner-images/issues/9262); you can find many discussions of this in Google, their decision to remove it broke a ton of builds). The weird thing is that our CI `.yaml` already has its own script to install conda, with a whole caching mechanism and such, so conda ought to be there even though GitHub Actions no longer installs it for us. But it isn't, apparently; I have no idea why, and debugging this kind of thing is very much not my forté. :-> If this sounds like something you can figure out, I'd appreciate the help! The issue is here: #454. But if not, no worries. Thank you very much for your help with this `make test` issue!

Issue #447, opened 2024-06-11
population name inscrutibly dissappears


petrelharp:
Over in https://github.com/tskit-dev/pyslim/issues/343 we ran into the following issue.
Consider this script:
```
initialize() {
setSeed(123);
    defineConstant("param_set_id", 1);
	defineConstant("NAnc", 17147);
	defineConstant("mAncNS", 0);
	defineConstant("NAncS", 411);
	defineConstant("NAncN", 209);
	
	// ----- TIME PARAMETERS -----
	defineConstant("start_splitAnc", 10); 
	defineConstant("today_delay", 58);
	defineConstant("Split_Sagit_delay", 741);
	defineConstant("Split_Sagit_gen", start_splitAnc + Split_Sagit_delay - today_delay);
	defineConstant("today_gen", Split_Sagit_gen + today_delay);

	// ----- TREE-SEQUENCE RECORDING ----- 
	initializeTreeSeq(); 
	
	// set the overall mutation rate map
	initializeMutationRate(0);
	initializeMutationType("m1", 0.5, "f", 0.0); // Neutral mutations
	initializeGenomicElementType("g1", m1, 1.0);
	initializeGenomicElement(g1, 0, 1000000);
	
	//constant recombination along the chromosome:
	initializeRecombinationRate((5.56*1e-8)*10);
}

1 early() {
	// Create the ancestral population
	sim.addSubpop("p1", NAnc);
       // sim.treeSeqOutput("unused.trees"); // <------ (*)
}

start_splitAnc early() {
	sim.addSubpopSplit("p2", NAncS, p1);
}

start_splitAnc+1 early() {
	//remove ancestral pop
	p1.setSubpopulationSize(0);
}

today_gen late() {
    sim.treeSeqOutput("orig_mod1.trees");
}
```
If we run this then (a) all the first generation individuals are in `p1` and are correctly retained in the tree sequence, but (b) population `p1` doesn't have its name still:
```
$ tskit populations orig_mod1.trees 
id	metadata
0	None
1	None
2	{'migration_records': [], 'name': 'p2', 'sex_ratio': 0.0, 'slim_id': 2}
```
However, if you output the tree sequence earlier, when `p1` still existed, by uncommenting the line marked `(*)`, then you get:
```
$ tskit populations orig_mod2.trees 
id	metadata
0	None
1	{'name': 'p1'}
2	{'migration_records': [], 'name': 'p2', 'sex_ratio': 0.0, 'slim_id': 2}
```

[...]

PR #459, opened 2024-07-20
tests for pop name retention

petrelharp:
As requested in #447. Currently this isn't very creative about what's tested; if you think of other weird scenarios we might add those.

### DIFF
--- a/treerec/tests/recipe_specs.py
+++ b/treerec/tests/recipe_specs.py
@@ -35,6 +35,8 @@ recipe_specs = {
     "test_____retain_and_remember_individuals.slim": {"marked_mutations": True, "individuals": True},
     "test_____retain_individuals_nonWF_unary.slim": {"marked_mutations": True, "individuals": True},
     "test_097 modeling nucleotides.slim":     {"mutations": True},
+    "test_____pop_names_pX.slim":		      {"mutations": True},
+    "test_____pop_names_nondefault.slim":     {"mutations": True},
 }
 
 
--- a/treerec/tests/test_recipes/test_____pop_names_nondefault.slim
+++ b/treerec/tests/test_recipes/test_____pop_names_nondefault.slim
@@ -1,0 +1,33 @@
+initialize() {
+
+	// TREE SEQUENCE TEST OUTPUT
+	initializeSLiMOptions(keepPedigrees=T); // necessary for pedigreeID
+	source("init.slim");
+	initializeTreeSeq(); 
+	
+	initializeMutationRate(0);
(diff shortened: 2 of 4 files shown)


bhaller:
Awesome.  I see that right now this is failing, as expected.  I'll fix the problem over in #447.  At that point, even if I push my (untested) fixes to the main branch, this PR is still based at the point before the fix, so it will remain red.  So... then I ask you to rebase this PR, and after rebasing it will go green?  Maybe I am over-thinking this, and should just optimistically merge this now and let the CI break, and then do my fix and see the CI go back to green?

petrelharp:
I was imagining you'd push your changes to *this* branch? To do that:
```
git remote add peter git@github.com:petrelharp/slim
git fetch peter
git checkout -b test_pop_names peter/test_pop_names
# <do edits; commit changes>
git push --set-upstream peter
```

bhaller:
> I was imagining you'd push your changes to _this_ branch?...

Oh, interesting!  I never knew I could push changes to someone else's PR; I mean, you're the owner of this PR, no?  I guess I have permission to do that because the PR is based on a branch (master branch of the SLiM repo) that I have permissions for?  Huh.  Learn a new thing about git every day.  :->

petrelharp:
Yep - I think whether or not you can push to my fork is a configurable option, but it's on by default.

petrelharp:
Otherwise, you could your own branch that's a copy of mine and then work on top of it:
```
git remote add peter git@github.com:petrelharp/slim
git fetch peter
git checkout -b my_pop_names peter/test_pop_names
# <do edits; commit changes>
git push --set-upstream origin
# <go open a new PR with your branch, called my_pop_names>
```

bhaller:
> Yep - I think whether or not you can push to my fork is a configurable option, but it's on by default.

Well, turns out it doesn't work at present:

```
bhaller@glass SLiM % git fetch peter                                    
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Never mind, I'll just let the tests go red for a bit, and merge this PR now.

petrelharp:
Hm - I see you couldn't push to it because this box wasn't checked:
Screenshot from 2024-07-21 07-48-47
... I thought it would be checked by default; I guess not here?

bhaller:
> Hm - I see you couldn't push to it because this box wasn't checked: Screenshot from 2024-07-21 07-48-47 ... I thought it would be checked by default; I guess not here?

Huh.  Where does that checkbox live?  Good to know for future reference.  I don't recall ever seeing it.

PR #464, opened 2024-08-20
Qt 6 shift

bhaller:
This PR adds the ability to compile with Qt6 rather than Qt5.  Seems to work on my local machine with Qt 6.7.2.  There may be a minimum Qt6 version that it works with, since Qt6 rolled out old functionality in stages, but with any reasonably modern Qt6 I think it will work.  I don't expect any difficulties cross-platform; changes were really quite minimal.  I am therefore going to optimistically merge this and see how it goes.


bhaller:
Note that this pushes the required Qt version from 5.9.5 to 5.15.0.  This may be an issue for users on very old platforms like Ubuntu 18.04 LTS.  For Qt 6 compatibility, APIs deprecated in the 5.15 series needed to be removed, so this is fairly unavoidable.  The solution is to update to at least 20.04 LTS, which I expect to work fine.

PR #483, opened 2024-11-11
Fix Windows CI (choose correct python executable)

### LINKED ISSUE #482: "no module named pytest" error in GitHub Actions for Windows
A new CI error from GitHub Actions on Windows:

C:\Users\runneradmin\miniconda3\envs\anaconda-client-env\python.exe: No module named pytest

Here, for example:

https://github.com/MesserLab/SLiM/actions/runs/11772475813/job/32787944593

@currocam you're good at figuring this sort of thing out; can I ask you to look at this?  I have no idea but the cause is, but this automated test build ran fine last week:

https://github.com/MesserLab/SLiM/actions/runs/11658388612

so I guess something must have changed in the runner setup in GitHub Actions...?

currocam:
This PR fixes #482 

As you already said in the issue, something must have changed since the last time the tests ran. Other people have very similar issues: 

https://github.com/conda-incubator/setup-miniconda/issues/371
https://github.com/conda-incubator/setup-miniconda/issues/355 

I haven't figured out *what exactly* has changed, but I think the issue was that the correct Python executable (the one for which we installed the pytest library) was not in PATH. You cannot ssh to the GitHub action (as far as I know) so it's extremely annoying to debug. I ended up going through their documentation and updating the code according to their "best practices". Eventually, it all worked out.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -119,10 +119,9 @@ jobs:
             msys2-devel
             mingw-w64-${{matrix.env}}-toolchain
             mingw-w64-${{matrix.env}}-cmake
-      - name: Setup Mambaforge
+      - name: Setup Miniforge3
         uses: conda-incubator/setup-miniconda@v3
         with:
-          miniforge-variant: Mambaforge
           miniforge-version: latest
           activate-environment: anaconda-client-env
           use-mamba: true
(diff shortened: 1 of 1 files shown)


bhaller:
Interesting.  I think you shifted us to mamba a while back, in https://github.com/MesserLab/SLiM/commit/cbc9e33ad19b8065229eb171997d21d0395581f7.  I don't recall what the motivation was; I think maybe (1) the CI was failing on recent macOS targets because they had discontinued something else, and/or (b) we wanted to get off of anaconda because they are threatening to charge money in future.  Maybe you can track down some discussion, associated with that commit?  There must've been a PR for that change, but I'm not sure how to go from a commit to the PR that the commit was in, in GitHub's UI.  :-O  Anyhow, thank you very much for this patch, @currocam.  If you believe it to be the correct fix, even in light of that previous commit, then I'll approve it since you know far better than I.  I just wanted to check with you about this first, to make sure.  :->

currocam:
Hi Ben,
I think this https://github.com/MesserLab/SLiM/pull/466 is the PR you are thinking of. In fact, using conda implies using the "default" channel, for which you are suppose to pay if you are a big organization. I don't think they are truly implemented it, but I prefer choosing the fully open-source "flavour". 

However, my truly original motivation to switch to mamba was that (1) it was the solver recommended in the documentation of the GitHub action plugin and (2) I was able to solve the MacOS issue first with mamba (https://github.com/MesserLab/SLiM/pull/457). 

However, I think this error was not related with using mamba or conda, but perhaps with the [latest automatic update](https://github.com/conda-incubator/setup-miniconda/commit/e2966bbc68d826b5925be032b44fad6efd41760c) in the version of the dependency the plugin we use uses to handle the cache. I think this PR is OK to merge.

PR #486, opened 2024-12-10
Multi-chromosome with tree sequences

bhaller:
This pull request is built on top of the multi-chromosome pull request, #470, which is still in progress.  This PR contains modifications to tree-sequence recording to make it compatible with multiple chromosomes, with one tree sequence per chromosome.  This is still very much a work in progress.  Several things are presently missing:

- the built-in python-based tests of tree-sequence recording broke with the multichromosome revisions, and have not yet been fixed
- documentation changes associated with these code changes have not yet been made
- writing of multi-chromosome tree sequences is implemented, but *all* writing out of tree sequences has been disabled; I want to get the other aspects of the design ironed out before I tackle that
- there are various bits that have been pended, marked with `FIXME MULTI_TREESEQ`
- the diffs are sometimes weird, because the underlying multichrom branch was kind of hacked up to keep things compiling and running as the multi-chromosome work progressed; the diffs in this PR are against that hacked-up Frankenstein's monster code, so to see what has really changed, you might compare against the current master branch version of SLiM instead – but then you'll get diffs for the HUGE number of changes in the multichrom branch, too
- the underlying multichromosome work is still in progress, and there are various things broken because of that.

Enjoy!


core/slim_globals.cpp
```diff
@@ -1647,17 +1647,37 @@ void SLiM_ConfigureContext(void)
...
+R"V0G0N({"$schema":"http://json-schema.org/schema#","additionalProperties":false,"codec":"struct","description":"SLiM schema for node metadata.","examples":[{"slim_id":123,"is_null":0}],"properties":{"slim_id":{"binaryFormat":"q","description":"The 'pedigree ID' of the haplosomes associated with this node in SLiM.","index":0,"type":"integer"},"is_null":{"binaryFormat":"B","description":"A vector of byte (uint8_t) values, with each bit representing whether the haplosome in the corresponding chromosome is a null haplosome (1) or not (0).","index":1,"type":"integer"}},"required":["slim_id","is_null"],"type":["object","null"]})V0G0N";
```
petrelharp, line 1680: minor point - the description should say what order the bits come in - e.g., does the least-significant bit in the first entry correspond to the first chromosome in the list, or what?

core/slim_globals.cpp
```diff
@@ -1647,17 +1647,37 @@ void SLiM_ConfigureContext(void)
...
+// the metadata here.  I think this is actually not allowed in JSON Schema, understandably.
+// The schema here provides 8 bits, so it suffices for up to an 8-chromosome model.  We can
+// have up to 256 chromosomes, so we would have up to 32 uint8_t bytes of is_null data here,
+// in that extreme case.  It is very ugly, but I think the solution might be to write out
+// a different version of the metadata schema depending on the number of bytes used.  In
+// other words, if 7 bytes of is_null data are needed (for 49-56 chromosomes), we'd write
+// out a version of the schema that specifies 7 bytes of is_null data, named is_null1 to
+// is_null7, I suppose.  This effectively puts the count into the schema itself.  Peter,
+// do you see a better option than this?  I'd really prefer not to add a count, since it's
+// completely redundant.
```
petrelharp, line 1678: Hm - well, looking at [the tskit metadata docs for binary encoding](https://tskit.dev/tskit/docs/latest/metadata.html#struct), I agree - arrays have to carry their length along with them. Hm - but, how about using a [string](https://tskit.dev/tskit/docs/latest/metadata.html#binaryformat)? The schema would still depend on the number of chromosomes, since you'd have like `"binaryFormat" : "10p"`, where `10` is the number of char's we need for that many chromosomes, but seems a lot nicer than some sequence of variables?

We could extend `tskit`'s metadata schema parsing to make this easier if necessary; but, we're still limited by the [`struct` module](https://docs.python.org/3/library/struct.html), and I'm not seeing a better option there than like `10p`.

petrelharp: Hm - actually the `p` specification encodes the length, so that doesn't help, but something like that (`10c`?) might?
Screenshot from 2024-12-20 12-01-46

petrelharp: Yes - I think my proposal is to do like `"binaryFormat" : "10c"` where `10` is how many bytes you want.

core/slim_globals.cpp
bhaller: aha, ok; wasn't sure how verbose to get

core/slim_globals.cpp
bhaller: OK – and write a different schema (`1c` vs. `10c` etc.) based on how many bytes are needed.  I'll have to review B vs. c, but isn't c a signed char, and B is a uint8?  Unsigned seems better; having the highest bit interpreted as a sign bit just makes things that much more bug-prone.

core/species.cpp
```diff
@@ -3387,6 +3514,49 @@ void Species::AboutToSplitSubpop(void)
 	community_.tree_seq_tick_offset_ += 0.00001;
 }
 
+void Species::CopySharedTablesIn(tsk_table_collection_t &p_tables)
```
petrelharp, line 3517: Maybe this should be called "LinkSharedTablesIn" or something other than "copy"?

petrelharp: Hm, or `LinkSharedTableColumns`?

petrelharp: Wait, I'm confused here - isn't `p_tables.nodes` just a [pointer to a NodeTable](https://tskit.dev/tskit/docs/latest/c-api.html#c.tsk_table_collection_t)? And the [NodeTable object](https://tskit.dev/tskit/docs/latest/c-api.html#c.tsk_node_table_t) has it's own `num_rows`? From the description it sounds like you're reassigning the pointers to the separate columns of the node table, but that's not what's happening here?

bhaller: Not sure whether that clarifies things sufficiently or not...

core/species.cpp
```diff
@@ -3854,126 +4026,405 @@ void Species::SimplifyTreeSequence(void)
...
 		
-		tsk_flags_t flags = TSK_SIMPLIFY_FILTER_SITES | TSK_SIMPLIFY_FILTER_INDIVIDUALS | TSK_SIMPLIFY_KEEP_INPUT_ROOTS;
+		// BCH 12/9/2024: Removing TSK_SIMPLIFY_FILTER_INDIVIDUALS here, because we now need to filter the individuals
+		// table ourselves after simplifying all the tree sequences (perhaps in parallel); see SimplifyAllTreeSequences().
+		tsk_flags_t flags = TSK_SIMPLIFY_FILTER_SITES | TSK_SIMPLIFY_KEEP_INPUT_ROOTS;
+		
+		// BCH 12/10/2024: This should still work, with our own node table filtering code.  As Jerome explains, "simplify
+		// will still keep the *edges* that are unary, and that's all that matters. The downstream filtering code you
+		// have just looks to see what nodes have references, and filters out those that are not used in any edges."
 		if (!retain_coalescent_only_) flags |= TSK_SIMPLIFY_KEEP_UNARY;
```
petrelharp, line 4044: I think this is supposed to be `TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS`; see #487.

core/species.cpp
```diff
@@ -3854,126 +4026,405 @@ void Species::SimplifyTreeSequence(void)
...
+	std::vector<tsk_id_t> samples;
 	
-	for (auto subpop_iter : population_.subpops_)
+	// BCH 7/27/2019: We now build a hash table containing all of the entries of remembered_nodes_,
+	// so that the find() operations in the loop below can be done in constant time instead of O(N) time.
+	// We need to be able to find out the index of an entry, in remembered_nodes_, once we have found it;
+	// that is what the mapped value provides, whereas the key value is the tsk_id_t we need to find below.
+	// We do all this inside a block so the map gets deallocated as soon as possible, to minimize footprint.
+	// BCH 12/9/2024: The point of all this kerfuffle is that an extant individual might also be a
+	// remembered individual, and we don't want to put it into the samples vector twice, I think.
```
petrelharp, line 4078: Maybe also that we need to put the samples first?

core/species.cpp
```diff
@@ -3854,126 +4026,405 @@ void Species::SimplifyTreeSequence(void)
...
+		// remap node references
+		for (Chromosome *chromosome : chromosomes_)
 		{
-			int64_t num_tracked = t.num_tracked_samples[root];
+			slim_chromosome_index_t chromosome_index = chromosome->Index();
+			TreeSeqInfo &chromosome_tsinfo = treeseq_[chromosome_index];
+			tsk_table_collection_t &chromosome_tables = chromosome_tsinfo.tables_;
+			tsk_id_t *edge_child = chromosome_tables.edges.child;
+			tsk_id_t *edge_parent = chromosome_tables.edges.parent;
+			tsk_size_t edges_num_rows = chromosome_tables.edges.num_rows;
```
petrelharp, line 4201: We also need to do `chromosome_tables.mutations.node`. I think we don't need to loop over `mutations.node` to check for which nodes to keep because `simplify` will have removed any mutations not ancestral to any samples; so the only mutations that we will still have that aren't on an edge are above either a root (which is parent to an edge, so will still be there) or above a sample (also still there). But we probably ought to check if `mutations_node[k]` is out of bounds for `node_id_map` and if so throw an internal error.

core/species.cpp
petrelharp: Ah, yep - sorry!

core/slim_globals.cpp
petrelharp: might as well be verbose enough to use the info!

core/slim_globals.cpp
petrelharp: That's the idea, you know better than I about the precise type.

core/species.cpp
```diff
@@ -3854,126 +4026,405 @@ void Species::SimplifyTreeSequence(void)
...
+		tsk_bool_t *keep_individuals = (tsk_bool_t *)calloc(num_individuals, sizeof(tsk_bool_t));	// note: cleared by calloc
+		tsk_id_t *individual_id_map = (tsk_id_t *)malloc(num_individuals * sizeof(tsk_id_t));
+		
+		if (!keep_individuals || !individual_id_map)
+			EIDOS_TERMINATION << "ERROR (Species::SimplifyAllTreeSequences): allocation failed; you may need to raise the memory limit for SLiM." << EidosTerminate(nullptr);
+		
+		// mark the individuals we want to keep: all individuals referenced by nodes; note that the node table is shared,
+		// so we only need to loop through that one shared node table that is kept by the main table collection
+		{
+			tsk_id_t *node_individual = main_tables.nodes.individual;
```
petrelharp, line 4260: minor style nit: elsewhere (eg in tskit) we'd call this `nodes_individual` so it's parallel to `nodel.individual` rather than `node_individual` (or `node_individuals`); nbd obvs tho

bhaller: Sure.  Please point this out anywhere you notice it, and I'll try to get into the habit.

core/species.cpp
bhaller: Perhaps you could magnify on this comment a bit.  I haven't delved into the code again yet, but I suspect I'm going to need clarification here when I do.  :->

core/species.cpp
```diff
@@ -4161,8 +4699,29 @@ void Species::RecordNewHaplosome(std::vector<slim_position_t> *p_breakpoints, Ha
...
 	right = (double)chromosome.last_position_+1;
 	tsk_id_t parent = (tsk_id_t) (polarity ? haplosome1TSKID : haplosome2TSKID);
-	int ret = tsk_edge_table_add_row(&tables_.edges, left, right, parent, offspringTSKID, NULL, 0);
+	int ret = tsk_edge_table_add_row(&tsinfo.tables_.edges, left, right, parent, offspringTSKID, NULL, 0);
 	if (ret < 0) handle_error("tsk_edge_table_add_row", ret);
+	
+	// BCH 12/10/2024: With the new metadata scheme for haplosome, we also need to fix the is_null_metadata if
+	// the new haplosome is a null haplosome *and* it belongs to a chromosome type where that is notable.  In
+	// the present design, that can only be chromosome types "A" and "H"; the other chromosome types do not
+	// allow deviation from the default null-haplosme configuration.
```
petrelharp, line 4708: IIUC - the node `is_null` metadata not record that, for instance, the second `X` of a male is null, since that goes without saying?

core/species.cpp
```diff
@@ -4395,7 +4967,7 @@ void Species::DerivedStatesFromAscii(tsk_table_collection_t *p_tables)
 										 binary_derived_state_offset.data(),
 										 mutations_copy.metadata,
 										 mutations_copy.metadata_offset);
-		if (ret < 0) handle_error("convert_from_ascii", ret);
+		if (ret < 0) handle_error("derived_to_ascii", ret);
```
petrelharp, line 4970: I think we're in DerivedStatesFromAscii here (not "to")?

bhaller: Ah, huh; I thought I was fixing a typo, but perhaps I introduced one, lol

core/species.cpp
```diff
@@ -5486,131 +6101,165 @@ void Species::WriteTreeSequence(std::string &p_recording_tree_path, bool p_simpl
...
+		// copied below; we will be modifying the tables, and don't want our modification to go into the original
+		// shared tables, which we are not allowed to change.
+		if (chromosome_index > 0)
+			CopySharedTablesIn(chromosome_tables);
+		
+		// Copy the table collection so that modifications we do for writing don't affect the original tables.
+		// Note that there's a lot of work below to clean up the individuals table and node table for saving.
+		// Those tables are shared.  We don't want to do this cleanup in the original tables, since that would
+		// modify our recording state I guess; but I think this cleanup will be the same for every chromosome,
+		// so technically we could do this work just once, I think (?), and share the processed tables across
```
petrelharp, line 6123: yes, this is definitely true, because we want the shared tables to be identical on output!

core/species.cpp
```diff
@@ -5486,131 +6101,165 @@ void Species::WriteTreeSequence(std::string &p_recording_tree_path, bool p_simpl
...
+		// all the chromosomes.  I've chosen not to pursue that idea, because I don't see a path to doing it
+		// without increasing the high-water mark for the memory usage of this code, which is very important
+		// to keep low.  Anyhow, maybe this is unimportant since it is only overhead at save time, and is
+		// probably not a hotspot.
```
petrelharp, line 6127: I agree, probably not a hotspot, but if you do, I don't see anywhere below where there will be an issue:
- sort doesn't touch nodes; and it reorders individuals based only on nodes;
- dedup_sites and compute_mutation_parents don't touch shared tables;
- alive indivs are added to the indiv table, but the same for all

So, it looks to me that if you just move the "add alive individuals" outside the loop over chromosomes, and copy tables over, everything should just work?

core/species.cpp
```diff
@@ -5623,11 +6272,24 @@ void Species::FreeTreeSequence()
...
 		// Free any tree-sequence recording stuff that has been allocated; called when Species is getting deallocated,
 		// and also when we're wiping the slate clean with something like readFromPopulationFile().
-		tsk_table_collection_free(&tables_);
-		tables_initialized_ = false;
+		bool first = true;
+		for (TreeSeqInfo &tsinfo : treeseq_)
+		{
+			// the node, individual, and population tables are shared; avoid doing a double free
+			// (I don't think any of the shared tables should be copied at this point anyway,
+			// though; maybe there should be an assert here to that effect?)
```
petrelharp, line 6280: looks good but FYI `tsk_table_collection_free` uses [`tsk_safe_free`](https://github.com/tskit-dev/tskit/blob/main/c/tskit/core.c#L726), which is "safe" because it won't try to free a NULL pointer

core/species.cpp
bhaller: Hmm.  I was thinking that this was just a performance issue – that we would save some (probably very small) amount of time by doing the work just once.  I figured the work is deterministic and will produce the same result every time, from the same data, and so "identical on output" would be true in any case.  But are you saying that you think that is not the case, and that it is therefore important to do the work just once, and not just for performance reasons?  If so, why?

bhaller: Ah, I see you write more on this below...

core/species.cpp
bhaller: Huh, OK, I'll look into that, thanks.

core/species.cpp
bhaller: Ah, true.  Jerome's code in the example had the same "avoid a double free", so I just copied his approach, but perhaps it is technically unnecessary.

bhaller: (and technically free() is safe to call with NULL anyway – https://stackoverflow.com/a/6084233/2752221)

petrelharp:
I've read it all reasonably carefully (but the validation code less carefully); no issues other than my comments! :tada:

Very nice!!

core/species.cpp
petrelharp: Nope - I agree that it should be deterministic, and if it is not then there is A Problem; but it seems like the code is cleaner if the work is done once? Maybe not?

core/species.cpp
petrelharp: The `mutation` table has a `node` column; here we are changing which node is which, so we need to also remap them, like `mutations_node[k] = node_id_map[mutations_node[k]]`.

You might also think we need to loop through the mutation table to add nodes that are referred to there to `keep_nodes`, but I don't think that's true - we should be able to `assert node_id_map[mutations_node[k]] >= 0`. (it'll be `-1` if the node has been removed)

petrelharp: Jerome's example didn't have mutations.

bhaller:
> I've read it all reasonably carefully (but the validation code less carefully); no issues other than my comments! 🎉
> 
> Very nice!!

Yay!  Thanks so much!  I'll try to put this to bed in the next day or two.  I may have a question or two for you, if you have time.  :->

core/species.cpp
petrelharp: oh p.s. also here edge*s*_child, etcetera

core/species.cpp
bhaller: Maybe.  I didn't see a clear/simple path to doing the work once.  I'll think about it more.

core/species.cpp
bhaller: OK, that last comment – what about edges_child?  And what's the etcetera?  You know this stuff way better than I do; if there are other references I need to be fixing up, please enumerate them exhaustively if you can!

petrelharp: sorry, I meant replace `edge_child` with `edges_child`, as below

core/slim_globals.cpp
bhaller: OK, I've done this.  I'm going to "resolve" these conversations as I go, so I can easily see what is left for me to do.  Resolving this now.

core/slim_globals.cpp
bhaller: Done, resolving.

core/species.cpp
bhaller: I've put a comment in the code quoting your comment here.  This is not related to multichrom at all, so I'm going to postpone dealing with it; one can of worms at a time.  Resolving this thread.

core/species.cpp
bhaller: I changed the comment to:

```
// BCH 12/9/2024: The point of all this kerfuffle with the lookup table is that an extant individual
// might also be a remembered individual, and we don't want to put it into the samples vector twice,
// I think; otherwise we could just throw the remembered nodes and extant individuals into `samples`
// with no lookup table complication.
```

I was trying to figure out, and then explain to myself, why `remembered_nodes_lookup` is needed, since that chunk of code is surprisingly complicated.  Hopefully my revised comment is clearer, and accommodates your feedback here.  Resolving this thread.

core/species.cpp
bhaller: I notice that tskit is actually not consistent about this.  I searched on `node_individual` to find places to clean up, and it found two places where tskit uses that instead of `nodes_individual` (at least in SLiM's outdated copy of the tskit code).  Just FYI.  I will fix it in SLiM nevertheless.  :->

bhaller: Done, resolving.

core/species.cpp
bhaller: fixed, resolving

core/species.cpp
bhaller: resolving, will not fix (I agree with myself that it would be hard to fix this cleanly, without increasing the high-water memory usage mark, and it is not important I don't think)

core/species.cpp
bhaller: I can't add alive individuals outside the loop over chromosomes, because then it would be doing it to the original tables (which we don't want to modify) rather than doing it on the copy of the table collection (which is made inside the loop, per chromosome).  So I think this really is trickier to fix that it might appear, and I think/hope the duplicated work is unimportant in the big picture.  Pursue this with me if you disagree.  For now, resolving this.

core/species.cpp
bhaller: Ah, I see why Jermome did this, and it is needed.  Each table collection its own copy of the pointers for columns.  So `tsk_safe_free()` doesn't protect this case.  You'd free the first table collection, would would zero out its column pointers, so if you tried to free _that_ table collection again, it would be OK – but the _other_ table collections still have non-zeroed pointers to the same memory that just got freed, and calling `tsk_safe_free()` on those stale pointers would still go boom.  Resolving this thread now (hopefully you're seeing these comments, even though I'm resolving!)

core/species.cpp
bhaller: OK, figured out what was meant by this, and I think I did it right.  :->  Resolving.

bhaller:
OK, this has cleared code review and seems to be working well enough for present purposes; it's ready to be merged into `multichrom` (but not `master`!)

Issue #498, opened 2025-03-25
GitHub tests failing for some mysterious reason


bhaller:
Hi @currocam.  I have need of your superpowers once again.

Tests just started failing on the branch I'm working on, `multichrom`, as evidenced by this:

https://github.com/MesserLab/SLiM/actions/runs/14052089256/job/39344856600

Annoyingly, it just says that the SLiM tests have failed; it doesn't show me the actual output from the failed tests, so I have no idea which tests, out of the tens of thousands that SLiM runs, are failing.  It says to use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely, but naturally it does say where or how to set those flags, nor can I find any help in Google on how to do so.  (People who write error messages like this ought to be court-martialled!  Not that I'm not guilty of it myself.  :->)

Normally, in this situation, the tests fail for me locally as well, so I can just see what the problem is on my own machine and fix it.  But – for the first time ever, I think – the tests are failing on GitHub, but *not* on my local machine, and so now I'm flummoxed.  How do I set those flags and see what's failing?  I hope you can help!  If you're busy, no worries; I'll try other folks who might know, but you always seem so good at this kind of thing!  If there's a way to just set "--output-on-failure" for all tests always, that would be good; I can't imagine why that isn't the default.  Who wants tests to fail and receive no output regarding what failed or why?

Thanks!

currocam:
Hi Ben, 
I believe the flags `—rerun-failed` and `—output-on-failure` are specific to CTest. We don’t directly invoke the `ctest` executable but rather use the `make test` command.  I think we do this for consistency reasons with build systems. 

Anyway, I believe in your working branch, you should be able to set the environment variable and execute the tests running `env CTEST_OUTPUT_ON_FAILURE=1 make test` instead of `make test`.

I intended to introduce an error, verify that the logs now include the test output, revert the error, and submit the PR. However, I believe the error I introduced didn’t occur. Could you take a look? I've never done any testing in  Cpp.

PR #499, opened 2025-03-25
Add verbose output when tests fail in CI

currocam:
Set up flags in CI so ctest prints an informative output to the log when a test fails (see #498)

I'm trying to follow this post:
https://stackoverflow.com/questions/5709914/using-cmake-how-do-i-get-verbose-output-from-ctest

I think the most sensible way is to set up those flags in the command when running CI, rather than overwriting the default behaviour of CMake, in case someone else than CI is using the Makefiles to build or test SLiM.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,7 +5,7 @@ on:
   push:
   pull_request:
   schedule:
-    - cron: '0 5 * * 1'
+    - cron: "0 5 * * 1"
 jobs:
   canceller:
     runs-on: ubuntu-latest
@@ -20,14 +20,14 @@ jobs:
         include:
           # Ubuntu with the oldest supported GCC
           # BCH 2/12/2025: Ubuntu 22.04 is being discontinued, shifting forward
-          - {os: ubuntu-22.04, gcc: 9, python: 3.9}
-          - {os: ubuntu-24.04, gcc: 12, python: 3.12}
+          - { os: ubuntu-22.04, gcc: 9, python: 3.9 }
+          - { os: ubuntu-24.04, gcc: 12, python: 3.12 }
           # macOS, oldest supported version
           # (macos-10.15, macos-11, and macos-12 were removed by GitHub)
-          - {os: macos-13, python: 3.9}
+          - { os: macos-13, python: 3.9 }
           # macOS, newest supported version
-          - {os: macos-15, python: 3.12}
-      
+          - { os: macos-15, python: 3.12 }
+
     runs-on: ${{ matrix.os }}
     steps:
       - name: Check out repository code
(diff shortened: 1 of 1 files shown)


bhaller:
Hi @currocam!  OK, great.  If I understand correctly, this should be ready to merge, and then the test errors should be visible in the CI automatically, which is awesome.  I take it the `—rerun-failed` flag isn't needed because we'll have this turned on all the time anyway, which is perfect.

By the way, where you now do:

```
          make -j 2 && \
          # Show any output from the test program whenever the test fails
          env CTEST_OUTPUT_ON_FAILURE=1 make test
``` 

I suspect that the `\` just joins the comment line onto the end of the first line, and then the third line remains a separate line...?  But I don't think it matters anyway; it seems to me that there was no particular reason for the two `make` commands to be joined by `&&` anyway, right?  Maybe I'm wrong and the `&&` is important, and maybe I'm wrong and the `\` skips over the comment and still joins the two `make` lines.  But I suspect you could delete the `&& \` and it would probably work just the same.  Anyhow, doesn't matter; I guess I'm just curious whether I've got a misunderstanding here, really.  :->  If you give me a thumbs-up I'll merge this, and then I'll get to find out why my tests are failing!  :->  Thanks!

bhaller:
Hmm, I'm gonna merge this now.  What's the worst that could happen?  :->  But I'm still curious about the `&& \` thing.  :->

bhaller:
It's working like a charm.  Figured out the weird test failure that didn't happen on my local machine.  The test failed only when run with command-line SLiM, not when run under SLiMgui, because there was an interaction there.  On my local machine I generally run the unit tests under SLiMgui, so I didn't see the failed tests.  Mystery solved.  :->  Thanks @currocam, I really appreciate the help!

currocam:
Great! I'm glad it works.  Mystery solved then :)

currocam:
> Hi @currocam! OK, great. If I understand correctly, this should be ready to merge, and then the test errors should be visible in the CI automatically, which is awesome. I take it the `—rerun-failed` flag isn't needed because we'll have this turned on all the time anyway, which is perfect.
> 
> By the way, where you now do:
> 
> ```
>           make -j 2 && \
>           # Show any output from the test program whenever the test fails
>           env CTEST_OUTPUT_ON_FAILURE=1 make test
> ```
> 
> I suspect that the `\` just joins the comment line onto the end of the first line, and then the third line remains a separate line...? But I don't think it matters anyway; it seems to me that there was no particular reason for the two `make` commands to be joined by `&&` anyway, right? Maybe I'm wrong and the `&&` is important, and maybe I'm wrong and the `\` skips over the comment and still joins the two `make` lines. But I suspect you could delete the `&& \` and it would probably work just the same. Anyhow, doesn't matter; I guess I'm just curious whether I've got a misunderstanding here, really. :-> If you give me a thumbs-up I'll merge this, and then I'll get to find out why my tests are failing! :-> Thanks!

About this, I didn't think about it. I think the && still refers to the third line. For example:
```
#!/bin/env bash

false && \
echo Hello

false && \
# Comment
echo Hello
```

As you said, I guess there's no reason to use `&&`. I think I did it to fail early if the build was not successful, but I don't think tests can pass (or even run) if the build is not successful.

bhaller:
Yep, that test shows that `\` jumps over the comment line in bash!  How funny.  It is not that way in C/C++:

#define FOO(y) y + \
y;
	FOO(10);

#define FOO(y) y + \
// foo!
y;
	FOO(10);

The former compiles; the latter gives an error, because the line `y;` is no longer part of the macro.  So I guess the preprocessor in bash removes comments first, and then joins lines with `\`, whereas C/C++ do it in the opposite order.  Learn a new thing every day.  :->

PR #502, opened 2025-03-31
python testing of multiple chromosomes

petrelharp:
WIP!


petrelharp:
And, good news! The tests seem to pass, locally. They're not testing anything with multiple chromosomes - this is just the old consistency checks; I need to add some new tests that exercise that side of things.

petrelharp:
Okay! Let's merge this; I'll work on other changes ina  separate PR?

petrelharp:
FYI on Windows CI we got
```
0s
Run cd windows_compat/gnulib && \
CMake Error at CMakeLists.txt:52 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```
... so I've added the last line to the cmake calls in `.github/workflows/tests.yml`.

PR #504, opened 2025-04-03
More tests: multichom treeseq

petrelharp:
Still (possibly?) TODO:
- [x] include a test that uses `H-` (deprecated, but still, should test?)
- [x] write a test that checks that chromosomal inheritance is happening as it should
- [x] write a test that checks whether `is_null` is as expected


bhaller:
I'm assuming that I shouldn't merge this until the to-do items are cleared away.  (They sound like a good idea to me!  :->)

PR #510, opened 2025-04-13
Update SLiM.spec to 5.0

bryce-carson:
Additionally utilize the new /archive/refs/tags/v%{version}.tar.gz path to the tagged version of SLiM rather the (apparently) no longer used /archive/v%{version}.tar.gz path.

This is the only work needed to prepare for the release of version five of SLiM for the Fedora-derived Linux-based operating systems.

**Remind me to enable internet access during the build on the day of the release so that OpenSUSE Tumbleweed is able to build, otherwise that chroot will fail to pull all packages from its upstream and that builder will fail. No other builds require internet access during the COPR build process.**

### DIFF
--- a/SLiM.spec
+++ b/SLiM.spec
@@ -25,13 +25,13 @@
 %endif
 
 Name:           SLiM
-Version:        4.3
+Version:        5.0
 Release:        1%{?dist}
 Summary:        an evolutionary simulation framework
 
 License:        GPLv3+
 URL:            https://messerlab.org/slim/
-Source0:        https://github.com/MesserLab/SLiM/archive/v%{version}.tar.gz
+Source0:        https://github.com/MesserLab/SLiM/archive/refs/tags/v%{version}.tar.gz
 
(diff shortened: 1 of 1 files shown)


bhaller:
Thanks Bryce!  I imagine on release day, **after** I have tagged the new version, I should then merge this PR when you tell me to?  I'll endeavor to remind you about the internet access.  :->

bryce-carson:
> Thanks Bryce!  I imagine on release day, **after** I have tagged the new version, I should then merge this PR when you tell me to?  I'll endeavor to remind you about the internet access.  :->

You can merge the PR now, actually. It won't hurt because you're merging it into the multichrom branch anyhow. When you've merged *that* branch into master, then tagged v5.0 is when I'll turn the cranks.

bhaller:
Aha, OK.  I'll merge it now, then.  Probably there will be several days between when `multichrom` gets merged and when the 5.0 release happens, but it sounds like that's fine.  :->

PR #511, opened 2025-04-13
Debug windows

petrelharp:
The problem here is that SLiM is not being allowed to delete a directory on Windows CI.

### DIFF
--- a/treerec/tests/init.slim
+++ b/treerec/tests/init.slim
@@ -30,10 +30,14 @@ function (void)initializeMarks(i$ num_marks)
    return;
 }
 
-function (void)saveTreeSeqFormats(void) {
+function (void)saveTreeSeq(void) {
    // create tree seq representation in various formats: this could overwrite any
    // existing ones output in the same cycle, but they should be identical
    // if done in the same cycle at the same point of the lifecycle
+   // Note that this should happen after other output functions, and only once;
+   // if not, then there will be errors in Windows CI where we get an error due
+   // to having to delete the directory when writing out the tree sequence file again.
+   setwdForGen();
    sim.treeSeqOutput("test_output.trees", simplify=T);
    sim.treeSeqOutput("test_output.unsimplified.trees", simplify=F);
 }
@@ -144,7 +148,6 @@ function (void)outputMutationResult(void) {
     * and make a copy, appending data from in the current cycle
     * * */
    baseDir = copyIntoNewWdForGen(mutationsFile); // copy the current mutationsFile
-   saveTreeSeqFormats();
    if(!exists("n_marks")) {
       // Normal output: append details of mutations
       for (pop in sim.subpopulations) {
@@ -182,7 +185,6 @@ function (void)outputMutationResult(void) {
 function (void)outputIndividuals(void) {
    // copy the existing individualsFile into a new subdir and append the current individuals
    baseDir = copyIntoNewWdForGen(individualsFile); // copy the current individualsFile
-   saveTreeSeqFormats();
    for (pop in sim.subpopulations) {
       addIndividuals(pop.individuals, "output");
    }
--- a/treerec/tests/test_recipes/test_000_ancestral_marks.slim
+++ b/treerec/tests/test_recipes/test_000_ancestral_marks.slim
@@ -35,6 +35,6 @@ early() {
 
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
-
+   saveTreeSeq();
 	sim.simulationFinished();
 }
--- a/treerec/tests/test_recipes/test_000_ongoing_muts.slim
+++ b/treerec/tests/test_recipes/test_000_ongoing_muts.slim
@@ -35,6 +35,6 @@ early() {
 
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
-
+   saveTreeSeq();
 	sim.simulationFinished();
 }
--- a/treerec/tests/test_recipes/test_000_sexual_WF.slim
+++ b/treerec/tests/test_recipes/test_000_sexual_WF.slim
@@ -24,6 +24,6 @@ initialize() {
 
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
-
+   saveTreeSeq();
 	sim.simulationFinished();
 }
--- a/treerec/tests/test_recipes/test_000_sexual_nonwf.slim
+++ b/treerec/tests/test_recipes/test_000_sexual_nonwf.slim
@@ -43,7 +43,7 @@ early() {
 
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
-
+   saveTreeSeq();
 	sim.simulationFinished();
 }
 
--- a/treerec/tests/test_recipes/test_000_simple_nonwf.slim
+++ b/treerec/tests/test_recipes/test_000_simple_nonwf.slim
@@ -42,7 +42,7 @@ early() {
 
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
-
+   saveTreeSeq();
 	sim.simulationFinished();
 }
 
--- a/treerec/tests/test_recipes/test_000_withdraw_indivs.slim
+++ b/treerec/tests/test_recipes/test_000_withdraw_indivs.slim
@@ -48,7 +48,7 @@ modifyChild() {
 
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
-
+   saveTreeSeq();
 	sim.simulationFinished();
 }
 
--- a/treerec/tests/test_recipes/test_002_quick_neutral.slim
+++ b/treerec/tests/test_recipes/test_002_quick_neutral.slim
@@ -22,4 +22,5 @@ initialize() {
 200 late() {
 	// TREE SEQUENCE TEST OUTPUT
 	outputMutationResult();
+   saveTreeSeq();
 }
--- a/treerec/tests/test_recipes/test_004 simple sexual A.slim
+++ b/treerec/tests/test_recipes/test_004 simple sexual A.slim
@@ -24,5 +24,6 @@ initialize() {
 100 late() { 
 	outputMutationResult();
 	outputIndividuals();
+   saveTreeSeq();
 	// cat("\nTotal mutation frequencies: " + sum(sim.mutationFrequencies(NULL)) + "\n"); 
 }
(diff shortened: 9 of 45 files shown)


petrelharp:
Finally got into the machine; here's the inscrutible full error message:
```
============================= test session starts =============================
platform win32 -- Python 3.9.13, pytest-8.3.5, pluggy-1.5.0
rootdir: D:\a\SLiM\SLiM\treerec\tests
plugins: xdist-3.6.1
collected 1 item

test_consistency.py E                                                    [100%]

=================================== ERRORS ====================================
_ ERROR at setup of TestWithMutations.test_mutation_consistency[test_____all_the_chromosome_types.slim] _

request = <SubRequest 'recipe' for <Function test_mutation_consistency[test_____all_the_chromosome_types.slim]>>
tmp_path_factory = TempPathFactory(_given_basetemp=None, _trace=<pluggy._tracing.TagTracerSub object at 0x000002012A7658B0>, _basetemp=WindowsPath('C:/msys64/tmp/pytest-of-runneradmin/pytest-2'), _retent
ion_count=3, _retention_policy='all')
worker_id = 'master'

    @pytest.fixture(scope="session", params=recipe_specs.keys())
    def recipe(request, tmp_path_factory, worker_id):
        """
        Return a dict containing information about the recipe results. Most logic in this
        fixture is to check whether we are running as a single proc, or as a worker.
        """
        recipe_name = request.param
        return_val = {"name": recipe_name}
        return_val.update(recipe_specs[recipe_name])
    
        if worker_id == "master":
            # not executing in with multiple workers
            run_dir = tmp_path_factory.getbasetemp() / recipe_name
        else:
            # get the temp directory shared by all workers, so that simulation files
            # are shared between workers
            run_dir = tmp_path_factory.getbasetemp().parent / recipe_name
        with FileLock(str(run_dir) + ".lock"):
            # lock until all simulations are done
            if run_dir.is_dir():
                return_val["results"] = make_result(run_dir)
            else:
                os.makedirs(run_dir, exist_ok=True)
>               return_val["results"] = run_slim(recipe_name, run_dir)

conftest.py:206: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
conftest.py:177: in run_slim
    retval = subprocess.call(cmd, cwd=script_dir, stderr=subprocess.STDOUT, stdout=out)
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\subprocess.py:349: in call
    with Popen(*popenargs, **kwargs) as p:
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\subprocess.py:951: in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Popen: returncode: None args: ['slim', '-s', '22', '-d', 'RUN_DIR="C:/msys6...>
args = 'slim -s 22 -d RUN_DIR=\\"C:/msys64/tmp/pytest-of-runneradmin/pytest-2/test_____all_the_chromosome_types.slim\\" D:\\a\\SLiM\\SLiM\\treerec\\tests\\test_recipes\\test_____all_the_chromosome_types.
slim'
executable = None, preexec_fn = None, close_fds = False, pass_fds = ()
cwd = 'D:\\a\\SLiM\\SLiM\\treerec\\tests', env = None
startupinfo = <subprocess.STARTUPINFO object at 0x000002012A7D8E20>
creationflags = 0, shell = False, p2cread = Handle(172), p2cwrite = -1
c2pread = -1, c2pwrite = Handle(892), errread = -1, errwrite = Handle(896)
unused_restore_signals = True, unused_gid = None, unused_gids = None
unused_uid = None, unused_umask = -1, unused_start_new_session = False

    def _execute_child(self, args, executable, preexec_fn, close_fds,
                       pass_fds, cwd, env,
                       startupinfo, creationflags, shell,
                       p2cread, p2cwrite,
                       c2pread, c2pwrite,
                       errread, errwrite,
                       unused_restore_signals,
                       unused_gid, unused_gids, unused_uid,
                       unused_umask,
                       unused_start_new_session):
        """Execute program (MS Windows version)"""
    
        assert not pass_fds, "pass_fds not supported on Windows."
    
        if isinstance(args, str):
            pass
        elif isinstance(args, bytes):
            if shell:
                raise TypeError('bytes args is not allowed on Windows')
            args = list2cmdline([args])
        elif isinstance(args, os.PathLike):
            if shell:
                raise TypeError('path-like args is not allowed when '
                                'shell is true')
            args = list2cmdline([args])
        else:
            args = list2cmdline(args)
    
        if executable is not None:
            executable = os.fsdecode(executable)
    
        # Process startup details
        if startupinfo is None:
            startupinfo = STARTUPINFO()
        else:
            # bpo-34044: Copy STARTUPINFO since it is modified above,
            # so the caller can reuse it multiple times.
            startupinfo = startupinfo.copy()
    
        use_std_handles = -1 not in (p2cread, c2pwrite, errwrite)
        if use_std_handles:
            startupinfo.dwFlags |= _winapi.STARTF_USESTDHANDLES
            startupinfo.hStdInput = p2cread
            startupinfo.hStdOutput = c2pwrite
            startupinfo.hStdError = errwrite
    
        attribute_list = startupinfo.lpAttributeList
        have_handle_list = bool(attribute_list and
                                "handle_list" in attribute_list and
                                attribute_list["handle_list"])
    
        # If we were given an handle_list or need to create one
        if have_handle_list or (use_std_handles and close_fds):
            if attribute_list is None:
                attribute_list = startupinfo.lpAttributeList = {}
            handle_list = attribute_list["handle_list"] = \
                list(attribute_list.get("handle_list", []))
    
            if use_std_handles:
                handle_list += [int(p2cread), int(c2pwrite), int(errwrite)]
    
            handle_list[:] = self._filter_handle_list(handle_list)
    
            if handle_list:
                if not close_fds:
                    warnings.warn("startupinfo.lpAttributeList['handle_list'] "
                                  "overriding close_fds", RuntimeWarning)
    
                # When using the handle_list we always request to inherit
                # handles but the only handles that will be inherited are
                # the ones in the handle_list
                close_fds = False
    
        if shell:
            startupinfo.dwFlags |= _winapi.STARTF_USESHOWWINDOW
            startupinfo.wShowWindow = _winapi.SW_HIDE
            comspec = os.environ.get("COMSPEC", "cmd.exe")
            args = '{} /c "{}"'.format (comspec, args)
    
        if cwd is not None:
            cwd = os.fsdecode(cwd)
    
        sys.audit("subprocess.Popen", executable, args, cwd, env)
    
        # Start the process
        try:
>           hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                                     # no special security
                                     None, None,
                                     int(not close_fds),
                                     creationflags,
                                     env,
                                     cwd,
                                     startupinfo)
E                                    FileNotFoundError: [WinError 2] The system cannot find the file specified

C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\subprocess.py:1420: FileNotFoundError
---------------------------- Captured stdout setup ----------------------------
Running ['slim', '-s', '22', '-d', 'RUN_DIR="C:/msys64/tmp/pytest-of-runneradmin/pytest-2/test_____all_the_chromosome_types.slim"', 'D:\\a\\SLiM\\SLiM\\treerec\\tests\\test_recipes\\test_____all_the_chro
mosome_types.slim'] in dir 'C:/msys64/tmp/pytest-of-runneradmin/pytest-2/test_____all_the_chromosome_types.slim', errors etc to 'SLiM_run_output.log'
=========================== short test summary info ===========================
ERROR test_consistency.py::TestWithMutations::test_mutation_consistency[test_____all_the_chromosome_types.slim] - FileNotFoundError: [WinError 2] The system cannot find the file specified
============================== 1 error in 0.17s ===============================
```
and there's nothing in `SLiM_run_output`.

petrelharp:
Deleting the `saveIndividuals` line restores this...

bhaller:
The "could not be removed" is super weird.  Frickin' Windows.  I guess for some reason the file gets locked and SLiM can't overwrite it.

bhaller:
Probably a file inside the trees archive folder is marked busy, or something like that?  It could conceivably be a SLiM bug, if SLiM is writing out to a file inside the trees archive and then never closing that open file reference, I suppose; but then I'd expect to get the same error on other platforms.  Hmm.

bhaller:
Is the script that triggers this error writing out a trees archive to the same path more than once?  Or is the existing directory at the same path a byproduct of a _previous_ trees test that used the same filename?

petrelharp:
Well, this test is saving out tree sequences twice to the same location. It's not the only recipe that is doing that, but it *is* the only recipe that's doing that with multiple chromosomes. So, the problem is that Windows lets us overwrite a previous file but not a previous directory. I'll adjust the code so that it doesn't do that, somehow.

bhaller:
Oh!  Yes.  I discovered this oddity myself, in a different context, while working with trees archives.  An oddity of OSes in general, not just Windows, is that they are happy to overwrite files with no error, but they all error if you try to overwrite a directory.  I wanted it to work properly to overwrite one trees archive with another, so I actually wrote a bunch of code in `Species::_CreateDirectoryForMultichromArchive()` that is intended to make that work.  Since just creating a new directory where the existing directory already is would fail at the OS level, it needs to loop through the contents and delete all the files inside, then remove the existing directory, and then the OS will let us create the new directory there.  HOWEVER, this is obviously a rather dangerous thing to do, so the code tries to be very careful.  It will only delete the existing directory if it "looks like" a trees archive; all the files in it have to end in `.trees`, and there can't be any subdirectories, and so forth.  It checks first and makes sure that the existing directory passes all tests, and if it does, then it proceeds to delete it and replace.  Huge PITA.

So, my guess is that Windows is putting something into that folder that causes the safety check to fail.  macOS does that, in fact; the Finder likes to spray files named `.DS_Store` all over your filesystem, and they would make the safety check fail, so the code actually explicitly allows files named `.DS_Store" to be deleted.  I'm guessing Windows must do a similar thing, and so there's another special-case spam exception that needs to be added to the code.  This is at line 7205 of species.cpp, in the current `multichrom` branch.  If you can identify what is inside the old trees archive that is causing the safety check to fail, I can add the exception.  (I could also make the code there actually print out the name of the offending file, in the error message, if that would be useful to you?)

Sorry I didn't twig to this a little earlier; as soon as you saw that it was about replacing one trees archive with another, I should've thought of this!  :-O

PR #515, opened 2025-04-20
small note in treeseq implementation

petrelharp:
This wasn't written down anywhere that I could find; I'm putting it in pyslim docs; but it should probably be in the manual somewhere too?

### DIFF
--- a/treerec/implementation.md
+++ b/treerec/implementation.md
@@ -211,21 +211,14 @@ in the tree sequence already (and the associated keys in the schema).
 We've not done that yet because making things exactly match seems like a pain,
 and no-one else is using the top-level metadata yet.
 
-### Top-level metadata:
+## Reading in tree sequences
 
-*Note: this is out of date.* Here's an example of the top-level metadata:
-```
-{
- "SLiM" : {
-     "model_type" : "WF",
-     "generation" : 123,
-     "file_version" : "0.5",
-     "spatial_dimensionality" : "xy",
-     "spatial_periodicity" : "x",
-     "separate_sexes" : true,
-     "nucleotide_based" : false
- }
-}
-```
+In addition to the various things mentioned above,
+To read in a tree sequence, SLiM requires that:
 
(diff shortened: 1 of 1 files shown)


bhaller:
I think perhaps this is at a level of detail that goes beyond what needs to be mentioned in the SLiM manual?  My focus there is more upon people who are trying to understand, and perhaps parse, output from SLiM.  Nobody except pyslim tries to write SLiM-compatible .trees files, and I don't think anybody has ever mentioned any aspiration to do so, either.  Do you really think this needs to be there?  I'll put it there if you think so, but I'm hesitant because I think it's just more detail than I anticipate any reader ever actually wanting to know.

petrelharp:
Well, I think that "format for input files" ought to be in manuals in general? Otherwise I agree with you. I was thinking in the 'technical docs' bit towards the end? Or perhaps writing this down in pyslim (which I have done) is good enough, though.

bhaller:
Decided you're right, worked it into an existing parenthetical statement in the doc:

> The first node for the individual is used for the first haplosomes of the chromosomes: the first autosomal haplosome, the first X haplosome, and the single Y haplosome.  The second node is used for the second haplosomes of the chromosomes: the second autosomal haplosome, and the second X haplosome.  (The second node never corresponds to a Y haplosome; it is unused.)  The first node therefore corresponds to all of the haplosomes inherited from the first parent, and the second node to those inherited from the second parent, for biparental crosses.  (Note that this is always the node table structure SLiM uses; even if you were to simulate only the Y, every individual would still reserve two nodes in the node table, the second of which would be unused.  In all cases, the two nodes corresponding to the haplosomes of a given individual must be adjacent in the node table, and sorted by their haplosome pedigree ID; the node for the first haplosome must come first in the node table.)

This doc all existed in 5.0, except that I added the final sentence.  Seems fine, and not confusing in this context, and since it's in a parenthetical users who aren't interested should feel free to ignore it.  :->  Thanks!

Issue #538, opened 2025-07-24
Is check for unique SLiM ID on haplosomes for not alive individuals when loading a tree sequence necessary?


petrelharp:
Currently on loading a tree sequence SLiM checks (a) if all loaded individuals have unique SLiM IDs; and (b) if *all* nodes have unique SLiM IDs. The latter behavior is minorly annoying for some use cases (involving merging of different tree sequences); and it's at least inconsistent. Seems like we should only check for alive nodes, and be mroe relaxed otherwise (since we don't do anything with them)?

bhaller:
@petrelharp can you upload a small test file here that currently produces this raise, and should not?  Thanks!

petrelharp:
Here is an example: [dup_nodes.txt](https://github.com/user-attachments/files/22197440/dup_nodes.txt)

And for reference here's the python script that produced it: 
```
import msprime, tskit, pyslim

ts = msprime.sim_ancestry(2)
ts = pyslim.annotate(ts, model_type='WF', tick=1)

t = ts.tables
t.nodes.clear()
for n in ts.nodes():
    if n.individual == 1:
        md = n.metadata
        md['slim_id'] = 0
        n.metadata = md
    t.nodes.append(n)
t.individuals.clear()
ind0, ind1 = list(ts.individuals())
t.individuals.append(ind0)
ind1.flags &= ~pyslim.INDIVIDUAL_ALIVE
t.individuals.append(ind1)
ots = t.tree_sequence()
ots.dump("dup_nodes.trees")
```

PR #554, opened 2025-09-07
Don't check for duplicate haplosome IDs

petrelharp:
As discussed in #538, the check that *no* haplosome IDs overlap is being overly restrictive for applciations involving parallel simulations. This moves that duplicate haplosome ID check to "only haplosomes in individuals", not all of them.

### DIFF
--- a/core/species.cpp
+++ b/core/species.cpp
@@ -8727,6 +8727,7 @@ void Species::__CreateSubpopulationsFromTabulation(std::unordered_map<slim_objec
 	// This could be done with a hash table, but I imagine that would be slower until the number of individuals becomes very large
 	// Also, I'm a bit nervous about putting a large number of consecutive integers into a hash table, re: edge-case performance
 	std::vector<slim_pedigreeid_t> pedigree_id_check;
+	std::vector<slim_haplosomeid_t> haplosome_id_check;
 	
 	gSLiM_next_pedigree_id = 0;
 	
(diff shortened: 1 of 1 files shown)


core/species.cpp
```diff
@@ -9812,13 +9817,12 @@ void Species::__CheckNodePedigreeIDs(__attribute__((unused)) EidosInterpreter *p
 	// could be nodes in the node table with haplosome pedigree IDs greater than those in use by individuals, in nonWF models.
 	// See https://github.com/MesserLab/SLiM/pull/420 for an example model that does this very easily.
 	
-	// Also, check for duplicate pedigree IDs, just in case.  __CreateSubpopulationsFromTabulation() does this for individual
-	// pedigree IDs; we do it for node pedigree IDs.  I decided to use a vector with std::sort() to check even though it is
-	// O(n log n), rather than a hash table for O(n), because I'm nervous about hitting a bad edge case with the hash table
-	// due to the nature of the values being inserted.  Shouldn't be a big deal in the grand scheme of things.
+	// Previously, we checked for duplicate pedigree IDs here as well, just in case.
+    // __CreateSubpopulationsFromTabulation() does this for haplosome IDs in
+    // living individuals already pedigree IDs; however, it was found to be
```
bhaller, line 9822: the comment text here is garbled and I'm not sure what it is intended to say; please fix

petrelharp: whoops; done

PR #555, opened 2025-09-08
reordered metadata schema and added tests for ordering; closes #543

### LINKED ISSUE #543: run metadata schemas through tskit.canonical_json( )
Next time there's an opportunity, we should adjust the metadata schemas so that they agree with the output of `tskit.canonical_json()`. This will avoid various headaches like this one https://github.com/tskit-dev/tskit/issues/3244.

petrelharp:
If the metadata schema isn't in the proper order, then the tests will print out JSON that can be inserted into the source code that is in the proper order, obtained by, for instance:
```
tskit.canonical_json(ts.tables.populations.metadata_schema.schema)
```
(and remembering that nodes has a placeholder that has to be inserted).


bhaller:
Did the tests not get added to this PR?  I see your changes to `slim_globals.cpp` here; that is the only file changed in this PR, so I'm a bit confused.  Also, remind me what you mean by "remembering that nodes has a placeholder that has to be inserted"?

PR #557, opened 2025-09-10
Update Eidos RNG on Windows to use Cryptography API; resolves issue #367

### LINKED ISSUE #367: Windows fix for RNG seeding needed
Hi @rdinnager!  This is one for you, when you have time.  It is not urgent but would be good to get in before the next version ships (which is not presently scheduled and probably not imminent).

In file `eidos_rng.cpp` at line 38 (at present), there is a function:

`static unsigned long int _Eidos_GenerateRNGSeed(void)`

This function is supposed to generate a new seed value for SLiM's random number generator from some operating system source of randomness.  For Linux and macOS, it now reads random bits from `/dev/random`, a cryptographically secure source of randomness (but very slow, which is why SLiM doesn't use it otherwise).  Unfortunately Windows does not have `/dev/random`, apparently, so for Windows it still uses the old scheme of hashing together things like the process ID and the current time in milliseconds, which is much less good – so much less good that there is perhaps some risk of multiple runs receiving the same seed value.

I'm not particularly worried about what solution you'd like to use, as long as it will give a different seed value every time SLiM is run, without depending upon any seed value supplied by the user on the command line etc.  This looks like one possibility:

https://en.wikipedia.org/wiki/Entropy_(computing)#Windows

but it doesn't really need to be cryptographically secure if that API is a pain to use.

Thanks!

chris-a-talbot:
This commit updates the Windows-specific implementation of _Eidos_GenerateRNGSeed() to use the BCryptGenRandom function from the modern Cryptography API: Next Generation (CNG). This provides a cryptographically secure source of randomness, bringing the Windows implementation up to par with the /dev/urandom method used on Linux and macOS.

The CMakeLists.txt file has also been updated to link the bcrypt library for the slim, eidos, and SLiMgui targets when building on Windows.

Resolves #367


bhaller:
Awesome, thanks for doing this!  It's a small issue, but has been hanging around for such a long time.  It will be lovely not to see it in the list of open issues any more.  :->

PR #561, opened 2025-09-10
Fix window visibility/positioning on Windows; resolves #546

### LINKED ISSUE #546: SLiMgui remembers window positions that are offscreen, making it unusable
If you use SLiMgui with a secondary monitor and move your modeling window to that screen, and then quit, disconnect the secondary monitor, and relaunch, SLiMgui can sometimes remember the position of the window and try to locate the window at a position that is now offscreen.  This seems to happen on Windows; possibly other platforms too, but I think I have always seen it with Windows users.  I tried to fix this bug in version 5.0, but the fix does not seem to work; the user sees the window flash on-screen momentarily but then get sent back to the off-screen position again, and there seems to be no way to move it on-screen (especially since Windows puts all of the app's menus inside the window, not at the top of the screen!).  Possibly the new on-screen coordinates that SLiMgui is trying to move the window to are bad, and so the OS decides to move it back again?  Unclear.  I need a Windows machine to debug this on.

chris-a-talbot:
SLiMgui, upon startup, attempts to open in the same position it was last in. However, in the case of multiple screens, this may end up being off-screen (if an external monitor was removed, for example). This commit ensures that, on startup, the main window first attempts to re-open at its last location; however, if that location is off-screen, it adjusts itself to somewhere visible. It also automatically updates its position in the event the available monitors are modified (removed, adjusted, etc.). In the event an app is open in fullscreen in the space needed to open, SLiMgui will attempt to overlay that fullscreen app. If this is not possible, it will trigger an alert on the taskbar/dock, notifying the user that the app has been opened in the background.

From what I could tell, the key problem was that `QtSLiMMakeWindowVisibleAndExposed` was not being called in `main.cpp` on startup. I think that one line may have been a complete fix. The remainder of the work is focused on improving the behavior of the windows in the context of display changes (or fullscreen apps), hopefully further avoiding visibility issues.

Resolves issue #546

### DIFF
--- a/QtSLiM/QtSLiMAppDelegate.cpp
+++ b/QtSLiM/QtSLiMAppDelegate.cpp
@@ -37,6 +37,8 @@
 #include "QtSLiMExtras.h"
 
 #include <QApplication>
+#include <QGuiApplication>
+#include <QScreen>
 #include <QOpenGLWidget>
 #include <QSurfaceFormat>
 #include <QMenu>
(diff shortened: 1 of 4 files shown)


bhaller:
LGTM, thanks!  I'll do some testing of it once it's merged, on macOS.

PR #565, opened 2025-09-24
update to tskit 1.0.0b2

petrelharp:
Probably we should not merge this and just wait for tskit 1.0.0; but we'll see if there's any issues.


petrelharp:
Well this works, but how about let's just leave this open and I'll use it to merge in the official release when it comes.

petrelharp:
Question - I think - resolved; merge away?

PR #588, opened 2025-12-15
SIMD optimization for dnorm()

andrewkern:
- Add SIMD-optimized dnorm() using SLEEF exp() for non-OpenMP builds
  - Handles both single and variable mu/sigma paths
  - Includes benchmark script in simd_benchmarks/
  - benchmark README updated with performance report


andrewkern:
okay @bhaller -- this is ready for your review

Issue #591, opened 2025-12-17
BUG: Student's T spatial kernel strength increases with distance instead of decreasing


andrewkern:
## description
The scalar tdist() function in SpatialKernel has an incorrect sign in the exponent, causing interaction strength to increase with distance instead of decrease. 

### expected behavior
Student's T kernel strength should decay with distance:
strength = fmax * pow(1 + (d/\tau)²/\nu, -(\nu+1)/2)
At distance 0.5 with fmax=5, \nu=3, \tau=0.5: strength should be 2.8125

### current behavior
  Strength increases with distance due to double-negative in exponent:
```C
  // spatial_kernel.h:77
  return max / pow(base, -(nu + 1.0) / 2.0);  // equals max * pow(base, (nu+1)/2)
```
  At distance 0.5: strength is 8.889 (exceeds fmax!)

### to reproduce
```C
  Reproduction:
  initialize() {
    initializeSLiMOptions(dimensionality='x');
    initializeInteractionType('i1', 'x', maxDistance=2.0);
  }
  1 early() { sim.addSubpop('p1', 2); }
  1 late() {
    p1.individuals[0].x = 0.0;
    p1.individuals[1].x = 0.5;
    i1.setInteractionFunction('t', 5.0, 3.0, 0.5);  // fmax=5, nu=3, tau=0.5
    i1.evaluate(p1);
    catn("strength at d=0.5: " + i1.strength(p1.individuals[0], p1.individuals[1]));
    catn("fmax: 5.0");
    // Output: strength=8.889 > fmax
  }
```

  ### cause:
  In core/spatial_kernel.h:77:
```C
  return max / pow(base, -(nu + 1.0) / 2.0);
```
  Dividing by pow(x, -n) equals multiplying by pow(x, n), so the negative exponent is cancelled out.

  ### fix:
  Change to either:
  - `return max * pow(base, -(nu + 1.0) / 2.0);` or
  - `return max / pow(base, (nu + 1.0) / 2.0);`

I will issue this patch in a subsequent PR @bhaller

PR #592, opened 2025-12-17
Add unit tests for spatial kernel SIMD implementations

andrewkern:
Adds comprehensive tests for spatial interaction kernel calculations:

  - Script-level tests verifying kernel formulas at known distances
  - SIMD vs scalar consistency tests comparing i1.strength() against i1.totalOfNeighborStrengths()
  - C++ level tests calling SIMD functions directly with 1000 random values

  All five kernel types covered: Linear, Exponential, Normal, Cauchy, Student's T.

  Note: Two tests fail, exposing a pre-existing bug in the scalar tdist() function (see #591)


andrewkern:
as expected-- two tests are failing in CI:
```md
FAILURE count: 2
SUCCESS count: 36837
```
this is good.

@bhaller, next I'll issue a separate PR with the patch on the original implementation of the Student's T kernel. Once that is merged we can rerun these tests and they should all go green

andrewkern:
okay this is all green! 🟢 

ready to merge @bhaller

PR #593, opened 2025-12-17
Fix Student's T kernel formula sign error in tdist()

### LINKED ISSUE #591: BUG: Student's T spatial kernel strength increases with distance instead of decreasing

andrewkern:
Fixes double-negative in exponent causing strength to increase with distance:

```C
  // Before (wrong): dividing by negative exponent = multiplying by positive
  return max / pow(base, -(nu + 1.0) / 2.0);

  // After (correct)
  return max * pow(base, -(nu + 1.0) / 2.0);
```

  Closes #591 
  
  Also should resolve CI test failures in #592

### DIFF
--- a/VERSIONS
+++ b/VERSIONS
@@ -26,6 +26,7 @@ development head (in the master branch):
 	add further SIMD optimizations for exp(), log(), log10(), log2() based on the SLEEF framework (https://github.com/shibatch/sleef), thanks to Andy Kern, https://github.com/MesserLab/SLiM/pull/587
 	SIMD optimization for dnorm() using the previous SLEEF work, thanks to Andy Kern, https://github.com/MesserLab/SLiM/pull/588
 	SIMD optimizations for spatial interaction strength calculations, thanks to Andy Kern, https://github.com/MesserLab/SLiM/pull/590
+	fix #593, Student's T distribution had a sign error in tdist(), thanks to Andy Kern, https://github.com/MesserLab/SLiM/issues/591
 
 
 version 5.1 (Eidos version 4.1):


bhaller:
This will also make it significantly faster, since multiplication is much faster than division.  Maybe this is why you saw such a large speedup for Student's T in your benchmarking, @andrewkern!  Your SIMD code was doing multiplication but the original code was (incorrectly) doing division?  :-O

PR #596, opened 2025-12-18
Replace manual 16x loop unrolling in _TallyMutationReferences_FAST with compiler-directed unrolling

andrewkern:
- Remove 30+ lines of hand-unrolled code in _TallyMutationReferences_FAST
  - GCC/MinGW: __attribute__((optimize("unroll-loops"))) on function
  - Clang: #pragma clang loop unroll(enable) on loop
  - Compilers auto-detect optimal unroll factor (~8x)
  - Targeted approach only affects this function, not all loops
  - Verified on Linux GCC, MinGW64, and Apple Clang; no performance regression


bhaller:
I like the second one more, because I don't want to specify the number of iterations to unroll.  But I'm surprised that Clang doesn't support that syntax that GCC supports; I thought that one of Clang's design stakes was that they support all the same options that GCC supports, so that they're a drop-in replacement for GCC?  But maybe that doesn't extend to every dusty corner like this.  :->  So sure, if solution 2 is the cleanest syntax we can get for unrolling in both compilers, with no specified number of unrolls, then let's do that.

core/population.cpp
```diff
@@ -7189,36 +7192,19 @@ void Population::_TallyMutationReferences_FAST_FromMutationRunUsage(bool p_clock
...
+				// loop bounds. This replaces previous manual 16x unrolling.
+				const MutationIndex * __restrict__ indices = mutrun->begin_pointer_const();
+				const int32_t mutrun_count = mutrun->size();
+				slim_refcount_t * __restrict__ refcounts = refcount_block_ptr;
+
+#if defined(__clang__)
+#pragma clang loop unroll(enable)
+#endif
+				for (int32_t i = 0; i < mutrun_count; ++i)
+					refcounts[indices[i]] += use_count;
```
bhaller: Hmm.  So, you've switched the loop to use an index `i` instead of the original pointer-range loop.  The claim made by the C++ STL folks is that pointer-range loops are actually faster (and that therefore C++ for loops using std::iterator are faster than old-style for loops that count from 0 to size-1).  So I'd suggest changing the style of this loop back, unless you have performance data indicating that this style is actually superior...?

bhaller:
Hi @andrewkern.  I think I'm going to back out this change.  I'm getting a warning on macOS, "Loop not unrolled: the optimizer was unable to perform the requested transformation".  Clang just seems to struggle with unrolling this loop, which is why I had it hand-unrolled in the first place.  I was optimistic that your changes would get it to co-operate, but it looks like that is not the case.  :-<  Sorry – I know you put some work into this!

Issue #594, opened 2025-12-17
allow pedigreeIDs to be accessed if treeSeq is enabled even if pedigree tracking is not


petrelharp:
This is a new version of #177; apologies.

Summary: currently, if we've done `initializeTreeSeq();` but not `initializeSLiMOptions(keepPedigrees=T);`, then doing `individual.pedigreeID` produces an error, saying "sorry, pedigree tracking features not available". The errors in question are [here](https://github.com/MesserLab/SLiM/blob/8834d1274ea3387dcc05fa381e5ed1f234dd270a/core/individual.cpp#L1330) and [here](https://github.com/MesserLab/SLiM/blob/8834d1274ea3387dcc05fa381e5ed1f234dd270a/core/individual.cpp#L1697).

My proposal is to change the current line:
```
			if (!subpopulation_->species_.PedigreesEnabledByUser())
```
to
```
			if (!subpopulation_->species_.PedigreesEnabledByUser() || treeSeqEnabled)  // <- not the correct thing but you understand
```
and then to say in the documentation for `pedigreeID`:
```
If pedigree tracking is turned on with initializeSLiMOptions(keepPedigrees=T), pedigreeID is a unique non-negative identifier for each individual in a simulation, never re-used throughout the duration of the simulation run.  If neither pedigree tracking or tree sequence recording are enabled, this property is unavailable.
```
Documentation and methods for parentPedigreeID and grandparentPedigreeID would be unchanged.

Rationale: currently, 
1. if tree sequence recording is turned on, then the pedigreeID is available, under the hood, and written out to the tree sequence;
2. but, turning on pedigree recording is required to have access to the pedigreeIDs in SLiM;
3. and, this is something you need if you want to do anything that matches things seen by SLiM to individuals in the output tree sequence (e.g., write out trait values to top-level metadata, or do something to a particular individual in SLiM that you identified in the tree sequence).
4. However, turning on pedigree recording (a) requires an extra, somewhat confusing, line of code; and (b) incurs additional computational burden, which according to  [this comment](https://github.com/MesserLab/SLiM/issues/177#issuecomment-829868845), can be substantial.

This is serious because if you want to do most anything with the output that's not just look at genotypes, then you need to write out extra information, and to do this you need to write the pedigreeIDs; maybe people aren't doing this much right now, but - I would argue - often they should be, or want to be.

[...]

PR #599, opened 2025-12-19
Add smooth_fast() method for SpatialMap

andrewkern:
Adds SIMD-optimized `smooth_fast()` method for SpatialMap convolution, preserving the original
  `smooth()` unchanged.

  - 2-6x speedup with SIMD (AVX2/SSE4.2/NEON)
  - 1.2-2.8x speedup even in scalar builds from loop reordering
  - 96 unit tests covering 1D/2D/3D, all kernel types, edge cases
  - Benchmark script in simd_benchmarks/
  
  ## benchmarking results
  
    SIMD Build (AVX2+FMA)

  | Test        | smooth() | smooth_fast() | Speedup |
  |-------------|----------|---------------|---------|
  | 1D 100 pts  | 0.006ms  | 0.002ms       | 2.96x   |
  | 1D 1000 pts | 0.131ms  | 0.031ms       | 4.22x   |
  | 2D 50x50    | 0.349ms  | 0.144ms       | 2.43x   |
  | 2D 100x100  | 5.615ms  | 1.202ms       | 4.67x   |
  | 2D 200x200  | 88.485ms | 15.035ms      | 5.89x   |
  | 3D 20x20x20 | 3.571ms  | 1.589ms       | 2.25x   |
  | 3D 30x30x30 | 12.394ms | 5.800ms       | 2.14x   |

  NO_SIMD Build (Scalar)

  | Test        | smooth() | smooth_fast() | Speedup |
  |-------------|----------|---------------|---------|
  | 1D 100 pts  | 0.006ms  | 0.003ms       | 2.24x   |
  | 1D 1000 pts | 0.097ms  | 0.084ms       | 1.16x   |
  | 2D 50x50    | 0.243ms  | 0.167ms       | 1.46x   |
  | 2D 100x100  | 3.595ms  | 2.336ms       | 1.54x   |
  | 2D 200x200  | 53.854ms | 39.838ms      | 1.35x   |
  | 3D 20x20x20 | 4.130ms  | 1.458ms       | 2.83x   |
  | 3D 30x30x30 | 14.161ms | 5.283ms       | 2.68x   |

### DIFF
--- a/core/spatial_map.cpp
+++ b/core/spatial_map.cpp
@@ -832,7 +832,7 @@ void SpatialMap::Convolve_S1(SpatialKernel &kernel)
 		
 		// Non-periodic: clamping above guarantees bounds, use SIMD
 		int64_t conv_a_first = a + kernel_a_first + kernel_a_offset;
-
+		
 		if (!periodic_a_)
 		{
 			int64_t count = kernel_a_last - kernel_a_first + 1;
@@ -848,7 +848,7 @@ void SpatialMap::Convolve_S1(SpatialKernel &kernel)
 			for (int64_t kernel_a = 0; kernel_a < kernel_dim_a; kernel_a++)
 			{
 				int64_t conv_a = a + kernel_a + kernel_a_offset;
-			
+				
 				// Wrap around periodic boundaries if out of range
 				if ((conv_a < 0) || (conv_a >= dim_a))
 				{
@@ -857,10 +857,10 @@ void SpatialMap::Convolve_S1(SpatialKernel &kernel)
 					while (conv_a >= dim_a)
 						conv_a -= (dim_a - 1);
 				}
-			
+				
 				double kernel_value = kernel_values[kernel_a] * coverage;
 				double pixel_value = values_[conv_a];
-			
+				
 				kernel_total += kernel_value;
 				conv_total += kernel_value * pixel_value;
 			}
@@ -871,7 +871,7 @@ void SpatialMap::Convolve_S1(SpatialKernel &kernel)
 	
 	TakeOverMallocedValues(new_values, 1, grid_size_);	// takes new_values from us
 }
-	
+
 // SIMD-accelerated convolution for 2D spatial maps
 // Uses vectorized dot products with loop reordering for contiguous memory access
 void SpatialMap::Convolve_S2(SpatialKernel &kernel)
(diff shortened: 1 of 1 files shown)


andrewkern:
no, there doesn't need to be a new function. this could replace the existing `smooth()`, but I wanted to make sure you were okay with the loop reordering so decided to put it in like this for a clean, easy to read diff. 

If you are okay with it, I can strip out the old function and change the name on this one

core/spatial_map.cpp
```diff
@@ -788,395 +788,382 @@ void SpatialMap::ColorForValue(double p_value, float *p_rgb_ptr)
...
-				while (conv_a < 0)
-					conv_a += (dim_a - 1);	// move -1 to dim - 2
-				while (conv_a >= dim_a)
-					conv_a -= (dim_a - 1);	// move dim to 1
+				int64_t conv_a = a + kernel_a + kernel_a_offset;
+
+				if ((conv_a < 0) || (conv_a >= dim_a))
+				{
+					if (!periodic_a_)
+						continue;
```
petrelharp: Codecov says this line isn't covered, suggesting there's no tests with periodic boundary conditions?

Here's what it looks like on Codecov, for @bhaller's interest:
Screenshot From 2025-12-19 10-59-49

andrewkern: yay codecov!

andrewkern: let me see if I can write a test to hit that...

VERSIONS
```diff
@@ -32,7 +32,8 @@ development head (in the master branch):
 	add SIMD acceleration for trig functions (sin(), cos(), tan(), asin(), acos(), atan(), atan2()), power functions (^ opperator, pow, SpatialMap.pow()), thanks to Andy Kern, https://github.com/MesserLab/SLiM/pull/595
 	update citations listed by citation()
 	CI workflow for calculating code coverage with Codecov, thanks to Andy Kern, https://github.com/MesserLab/SLiM/pull/601
-	fix #594, the pedigreeID and haplosomePedigreeID properties should be accessible (without other predigree-tracking features) in tree-seq models, even without pedigree tracking being enabled
+	fix #594, pedigreeID property should be accessible (without other predigree-tracking features) in tree-seq models
```
bhaller: incorrect merge; please take the new comment, "fix #594, the pedigreeID and haplosomePedigreeID properties should be accessible (without other predigree-tracking features) in tree-seq models, even without pedigree tracking being enabled"

core/spatial_map.cpp
petrelharp: wait that says that it's not hitting the *not* periodic case?

core/spatial_map.cpp
```diff
@@ -788,395 +788,382 @@ void SpatialMap::ColorForValue(double p_value, float *p_rgb_ptr)
...
 					}
 				}
-				
+
 				*(new_values_ptr++) = ((kernel_total > 0) ? (conv_total / kernel_total) : 0);
 			}
 		}
 	}
-	
-#if 0
```
bhaller: Hmm, why remove this `#if` 0 code?  It performs a useful self-check that should still work, right?  (Maybe it would need to compare with a tolerance now?  You don't need to fix it to work correctly; maybe just put a comment in it saying "needs to check with tolerance now due to SIMD"?

bhaller (commented):
Hi Andy!  Here's a quick review, hopefully helpful.

core/spatial_map.cpp
```diff
@@ -788,395 +788,382 @@ void SpatialMap::ColorForValue(double p_value, float *p_rgb_ptr)
...
 	int64_t dim_a = grid_size_[0];
 	double *new_values = (double *)malloc(dim_a * sizeof(double));
-	
+
 	if (!new_values)
 		EIDOS_TERMINATION << "ERROR (SpatialMap::Convolve_S1): allocation failed; you may need to raise the memory limit for SLiM." << EidosTerminate(nullptr);
-	
-	// this assumes the kernel's dimensions are symmetrical around its center, and relies on rounding (which is guaranteed)
+
 	int64_t kernel_a_offset = -(kernel_dim_a / 2);
```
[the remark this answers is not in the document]
bhaller: IIRC `kernel_a_offset` is negative because the kernel extends to the left and above the focal central point as the kernel is being applied.  It is the offset that needs to be added to translate from "inside the kernel" coordinates to spatial-map coordinates, I think?  Or something like that; it has been quite a while since I looked at this code.  And yes, the code relies on integer rounding, IIRC.  The logic is correct, I think?  Or at least the original logic prior to this PR was correct, I think, and this PR compared as equal to the results of the old code?  I'm relying on that fact rather than trying to review this line by line.  :->

core/spatial_map.cpp
```diff
@@ -788,395 +788,382 @@ void SpatialMap::ColorForValue(double p_value, float *p_rgb_ptr)
...
+// SIMD-accelerated convolution for 1D spatial maps
+// Uses vectorized dot products with loop reordering for contiguous memory access
 void SpatialMap::Convolve_S1(SpatialKernel &kernel)
 {
 	if (spatiality_ != 1)
 		EIDOS_TERMINATION << "ERROR (SpatialMap::Convolve_S1): (internal error) map spatiality 1 required." << EidosTerminate();
 	if (kernel.dimensionality_ != 1)
 		EIDOS_TERMINATION << "ERROR (SpatialMap::Convolve_S1): (internal error) kernel dimensionality 1 required." << EidosTerminate();
-	
+
```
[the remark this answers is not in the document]
andrewkern: yep... sorry. this difference between out editors has been getting me all week

core/spatial_map.cpp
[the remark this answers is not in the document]
andrewkern: yes this PR yields equivalent results to what existed perviously -- this part isn't new code

andrewkern:
i still have white space clean up to do... i incorporated a bunch of the review comments at this point however.

PR #601, opened 2025-12-19
fix up codecov runner

andrewkern:


andrewkern:
@bhaller -- this is working now! I'm adding a new file to the repo called `codecov.yml` that sits at the project root and sets thresholds/behavior for the auto-generated codecov check. 

To start out it will report a coverage failure if a PR drops coverage by >1%. We may want to change that down the line

PR #630, opened 2026-05-21
Confine SIMD code to runtime-dispatched tiers (fixes #628)

### LINKED ISSUE #628: SIMD/SLEEF optimizations cause crashes for old hardware
@andrewkern we weren't expecting this to come up, but here it is.  :->  From https://groups.google.com/g/slim-discuss/c/S4DF_5vh-KE/m/X-NJPf7CAAAJ:

> I have been running sims on a distributed cluster and have noticed crashes on jobs that were specifically sent to clusters with older hardware. Of those, most exit with some kind of opaque SIGILL error.  I didn't notice this with v5.0 (haven't tried v5.1). My hypothesis is that the updated vector operations are not compatible with x86_64-v2 hardware. Those 2012 cpus are pretty old now, but there are still plenty of HPCs that still have them in use! I was just wondering if this is a known thing on your end or if there is something I could fix on my end. Either way I was hoping I could get some clarity on hardware requirements from you if possible.

For now, users on hardware this old can build SLiM themselves with `-D USE_SIMD=OFF` in their `cmake` build command, IIRC.  But given that this has actually bitten a user, I guess it'd be good to get a solution to this that doesn't require the user to set build flags manually.  How difficult is that to detect at compile time?  There must be a way to do this within CMake...

andrewkern:
**Problem.** SLiM was built with -mavx2 -mfma applied project-wide, so the compiler emitted AVX2/FMA throughout the whole binary — not just the SIMD kernels. Pre-Haswell x86_64 CPUs (no AVX2) crashed with SIGILL (see issue #628).

  **Fix.** The kernels formerly in `eidos_simd.h` move to `eidos_simd_impl.h`, a tier-parameterized body compiled once per instruction-set tier — scalar, SSE4.2, AVX2+FMA (x86_64), NEON (ARM64).
 
Only the per-tier `.cpp` files get ISA flags; everything else, including the dispatcher, builds at the baseline x86_64 ABI. `Eidos_SIMD_Init()`  checks the CPU with `__builtin_cpu_supports()` and points the Eidos_SIMD function pointers at the fastest supported tier.

One binary should now correct on any x86_64 CPU — baseline contains no AVX2/SSE4.2, tier code runs only after the CPU is confirmed to support it. USE_SIMD is now ON/OFF (OFF and MSVC build scalar only).

  Verification.
  - -testEidos (7464) and -testSLiM (36853) pass.
  - Object-file inspection: AVX2/FMA only in eidos_simd_avx2.cpp.o, SSE4.2-only in eidos_simd_sse42.cpp.o, every other .o baseline-clean.
  - All three x86 tiers exercised at runtime (AVX2 / SSE4.2 / scalar) — all pass.
  - USE_SIMD=OFF build: zero AVX2/FMA, tests pass.

**Things still needed!**
  -  Xcode project. SLiM.xcodeproj needs updating before the macOS build will work:
  - Add the 6 new files — eidos_simd.cpp, eidos_simd_scalar.cpp, eidos_simd_sse42.cpp, eidos_simd_avx2.cpp, eidos_simd_neon.cpp, and the header eidos_simd_impl.h — and add the 5 .cpp files to the Compile Sources phase of each Eidos/SLiM target.
  - Set per-file flags in the Compile Sources build phase: -mavx2 -mfma on eidos_simd_avx2.cpp, -msse4.2 on eidos_simd_sse42.cpp. Without the flags that target I think it will fail to compile (AVX2 intrinsics with no -mavx2)
  - The qmake build is scalar-only (EIDOS_SUPPRESS_SIMD_DISPATCH) i think?

### DIFF
--- a/eidos/eidos_simd.cpp
+++ b/eidos/eidos_simd.cpp
@@ -2,7 +2,7 @@
 //  eidos_simd.cpp
 //  Eidos
 //
-//  Created by Andrew Kern on 5/21/2026.
+//  Created by Andrew Kern in 2026.
 //  Copyright (c) 2024-2025 Benjamin C. Haller.  All rights reserved.
 //	A product of the Messer Lab, http://messerlab.org/slim/
 //
--- a/eidos/eidos_simd.cpp
+++ b/eidos/eidos_simd.cpp
@@ -27,6 +27,8 @@
 
 #include "eidos_simd.h"
 
+#include <cstring>
+
 
 // The public kernel pointers.  They are statically initialized to the scalar
 // tier so that a call is well-defined even if it somehow happens before
@@ -44,43 +46,70 @@ enum class Eidos_SIMD_Tier { kScalar, kSSE42, kAVX2_FMA, kNEON };
 static Eidos_SIMD_Tier sActiveTier = Eidos_SIMD_Tier::kScalar;
 
 
-void Eidos_SIMD_Init(void)
+bool Eidos_SIMD_SelectTier(const char *tier_name)
 {
-	static bool initialized = false;
-	
-	if (initialized)
-		return;
-	initialized = true;
-	
+	// The scalar tier is built on every platform and always available.
+	if (std::strcmp(tier_name, "scalar") == 0)
+	{
+		Eidos_SIMD_Fill_scalar();
+		sActiveTier = Eidos_SIMD_Tier::kScalar;
+		return true;
+	}
+
 #if EIDOS_SIMD_DISPATCH_X86
 	// __builtin_cpu_supports() reads CPUID; it is available on GCC and Clang
 	// for x86 and works regardless of the flags this file was compiled with.
 	// AVX2 and FMA shipped together (Haswell), but we require both explicitly
 	// since the AVX2 tier and SLEEF both use FMA instructions.
-	if (__builtin_cpu_supports("avx2") && __builtin_cpu_supports("fma"))
+	if (std::strcmp(tier_name, "AVX2+FMA") == 0)
 	{
+		if (!(__builtin_cpu_supports("avx2") && __builtin_cpu_supports("fma")))
+			return false;
 		Eidos_SIMD_Fill_avx2();
 		sActiveTier = Eidos_SIMD_Tier::kAVX2_FMA;
-		return;
+		return true;
 	}
-	if (__builtin_cpu_supports("sse4.2"))
+	if (std::strcmp(tier_name, "SSE4.2") == 0)
 	{
+		if (!__builtin_cpu_supports("sse4.2"))
+			return false;
 		Eidos_SIMD_Fill_sse42();
 		sActiveTier = Eidos_SIMD_Tier::kSSE42;
-		return;
+		return true;
 	}
 #endif
-	
+
 #if EIDOS_SIMD_DISPATCH_ARM
-	// NEON is baseline on every ARM64 CPU, so no detection is needed.
-	Eidos_SIMD_Fill_neon();
-	sActiveTier = Eidos_SIMD_Tier::kNEON;
-#else
-	// The scalar tier: the fallback for pre-AVX2/pre-SSE4.2 x86, unknown
-	// architectures, MSVC, and USE_SIMD=OFF builds.  It runs on any CPU.
-	Eidos_SIMD_Fill_scalar();
-	sActiveTier = Eidos_SIMD_Tier::kScalar;
+	// NEON is baseline on every ARM64 CPU, so it is always available here.
+	if (std::strcmp(tier_name, "NEON") == 0)
+	{
+		Eidos_SIMD_Fill_neon();
+		sActiveTier = Eidos_SIMD_Tier::kNEON;
+		return true;
+	}
 #endif
+
+	return false;
+}
+
+void Eidos_SIMD_Init(void)
+{
+	// Install the fastest tier the CPU supports. This is idempotent: calling it
+	// again re-runs detection and re-installs the same tier, which is how the
+	// SIMD self-tests restore normal dispatch after cycling through every tier.
+#if EIDOS_SIMD_DISPATCH_X86
+	if (Eidos_SIMD_SelectTier("AVX2+FMA"))
+		return;
+	if (Eidos_SIMD_SelectTier("SSE4.2"))
+		return;
+#endif
+#if EIDOS_SIMD_DISPATCH_ARM
+	if (Eidos_SIMD_SelectTier("NEON"))
+		return;
+#endif
+	// Fallback for pre-AVX2/pre-SSE4.2 x86, unknown architectures, MSVC, and
+	// USE_SIMD=OFF builds: the scalar tier, which runs on any CPU.
+	Eidos_SIMD_SelectTier("scalar");
 }
 
 const char *Eidos_SIMD_ActiveTierName(void)
--- a/eidos/eidos_simd.cpp
+++ b/eidos/eidos_simd.cpp
@@ -2,8 +2,8 @@
 //  eidos_simd.cpp
 //  Eidos
 //
-//  Created by Andrew Kern in 2026.
-//  Copyright (c) 2024-2025 Benjamin C. Haller.  All rights reserved.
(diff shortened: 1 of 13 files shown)


andrewkern:
grrr... some of the windows tests are failing... working on it

andrewkern:
okay @bhaller -- this is ready for review

eidos/eidos_simd.cpp
```diff
@@ -0,0 +1,131 @@
+//
+//  eidos_simd.cpp
+//  Eidos
+//
+//  Created by Andrew Kern on 5/21/2026.
+//  Copyright (c) 2024-2025 Benjamin C. Haller.  All rights reserved.
```
bhaller: fix to 2026 only, please

eidos/eidos_simd.cpp
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+#if EIDOS_SIMD_DISPATCH_X86
+	// __builtin_cpu_supports() reads CPUID; it is available on GCC and Clang
+	// for x86 and works regardless of the flags this file was compiled with.
+	// AVX2 and FMA shipped together (Haswell), but we require both explicitly
+	// since the AVX2 tier and SLEEF both use FMA instructions.
+	if (std::strcmp(tier_name, "AVX2+FMA") == 0)
+	{
+		if (!(__builtin_cpu_supports("avx2") && __builtin_cpu_supports("fma")))
```
bhaller, line 66: I'm worried about compilers that don't support `__builtin_cpu_supports`.  I guess this is on GCC 4.8+ and Clang 9+.  Clang 9 was only released in 2019, so there might be older versions floating around pretty commonly.  GCC 4.8+ is fine, though; that's the first GCC version that supported C++11, so it is required anyway.  (Clang supported C++11 in version 3.3, so suddenly requiring Clang 9 is a BIG jump forward in requirements.)  Older GCC/Clang, and other compilers like MSVC and Intel, will not have `__builtin_cpu_supports`, though.  So I'm worried that we're substituting a common problem (compilers not supporting __builtin_cpu_supports) in place of a rare problem (hardware so old that this solution is needed in the first place).

I see that Google has a more general solution to this sort of problem: https://github.com/google/cpu_features.  It is not simple, and would complexify SLiM's build significantly; it is not a single-header solution, unfortunately, it would require setting up a CMake subdirectory and stuff.  Doubtless that would be a PITA to get working.  But maybe it is what we have to do?  Or we could detect compilers that don't have `__builtin_cpu_supports`, based on not-GCC, not-Clang, or old-GCC-or-Clang, and fall back to scalar while emitting a warning, or some such.  Ugh.  Thoughts?

eidos/eidos_simd.cpp
bhaller: and likewise in other files; I notice this is a repeating problem, due to copy-paste...

eidos/eidos_test_functions_math.cpp
```diff
@@ -1530,6 +1552,179 @@ void _RunSIMDMathTests(void)
...
+	// SSE4.2, AVX2+FMA, NEON); normal execution only runs the single tier the
+	// CPU selected at startup.  To test, and cover, every tier's kernels, cycle
+	// through all tiers available on this CPU, then restore the best one.
+	static const char * const tier_names[] = { "scalar", "SSE4.2", "AVX2+FMA", "NEON" };
+
+	for (const char * const tier_name : tier_names)
+	{
+		if (!Eidos_SIMD_SelectTier(tier_name))
+			continue;
+		std::cout << "Testing SIMD kernels: " << Eidos_SIMD_ActiveTierName() << " tier";
```
bhaller: It looks like this output is generated whenever the self-tests are run?  Normally the self-tests only print the "SUCCESS" message, unless something goes wrong; they're not intended to be verbose in the success case.  But maybe it makes sense to do this output, so the user can see what is getting tested.  I'm not sure, hmm.

bhaller: I suppose I'll try it and see how it feels, and I can always disable this output later on if I decide against it.  But feel free to comment if you have an opinion.

bhaller (commented):
Hi @andrewkern!  Sorry for the slow review, things got busy.  :-O  A couple concerns here, some trivial, some bigger.  See what you think.

eidos/eidos_simd.cpp
andrewkern: ack sorry about this. will fix that.

bhaller: No worries, a nit.  :->

eidos/eidos_simd.cpp
andrewkern: Hmmm. A few thoughts.
1. I think the google thing is overkill 
2. MSVC and Intel cc are already handled. EIDOS_SIMD_DISPATCH_X86 is gated on `(defined(__GNUC__) || defined(__clang__))`, so on MSVC/Intel the entire dispatcher is `#if`'d out and the scalar path is the only one
3. Maybe old CLANG versions (or apple's version) would be an issue?

I think this can all be handled with a CMake check like

```CMake
  include(CheckCXXSourceCompiles)
  check_cxx_source_compiles(
      "int main() { return __builtin_cpu_supports(\"avx2\") ? 1 : 0; }"
      EIDOS_SIMD_HAS_BUILTIN_CPU_SUPPORTS)
  if(NOT EIDOS_SIMD_HAS_BUILTIN_CPU_SUPPORTS)
      add_compile_definitions(EIDOS_SUPPRESS_SIMD_DISPATCH=1)
      message(STATUS "SIMD: compiler does not support __builtin_cpu_supports; falling back to scalar")
  endif()
  ```
  
  Want me to add that?

eidos/eidos_test_functions_math.cpp
andrewkern: We could collapse this to a one line summary -- I think we want something here though?

eidos/eidos_simd.cpp
andrewkern: wait just to be clear -- do you want the copyright date switched or my date just to be `Created by Andrew Kern in 2026`?

eidos/eidos_simd.cpp
bhaller: Yes, I think so.  Of course then we have to worry about what version of CMake added CheckCXXSourceCompiles.  This sort of thing sometimes feels like an infinite regress.  I have never figured out how to find out the first version of CMake that added a given feature; they don't make it easy to tell and even Google seems to have no idea.  But I'd say add this, let's not worry about the CMake version.  If it causes problems for someone, they can always turn off SIMD altogether for their build, and this check would be inside the SIMD stuff, so that would fix the problem if their CMake is that old.  Anyway, updating to a newer CMake version is easier than updating to new hardware, so this is certainly a step forward overall.  :->  Thanks!

eidos/eidos_test_functions_math.cpp
bhaller: OK, a one-line summary seems like a good idea...?

eidos/eidos_simd.cpp
bhaller: The "Created by" was fine as it was.  The copyright was the problem – it can't say that it was copyright in a year before it even existed.  :->  So the copyrights on these SIMD files should say "Copyright (c) 2026 Benjamin C. Haller.  All rights reserved."  (In giving this code to SLiM, you implicitly give the copyright to me, I hope you don't mind that.  Somebody needs to hold the copyright, and since I need control over the repo, that ought to be me, since there's no foundation/trust/whatever.  :->)

andrewkern: of course!

andrewkern:
okay @bhaller -- ready for your review

bhaller:
Hi @andrewkern.  Sorry for the slow reviews, on vacation now.  This looks good to me.  For Xcode, I guess I should define `EIDOS_SUPPRESS_SIMD_DISPATCH` as the `qmake` build does?  I think that's fine, for now...?  Or is the default `EIDOS_SUPPRESS_SIMD_DISPATCH` anyway?  I.e., what happens for a non-CMake build that simply doesn't define anything SIMD-related?

I'll merge this soon, I just want to fix up the Xcode build at the same time.  :->

PR #639, opened 2026-07-08
boolean not bool in schema types

petrelharp:
The current-in-`multitrait` metadata schemas fail to validate on [tskit's metadata metaschema](https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/tskit/metadata_schema.schema.json#L29) (and hence the tree sequences fail to load), because the `tag` types are `bool`, when they should be "`boolean`". This fixes that problem.

### DIFF
--- a/core/slim_globals.cpp
+++ b/core/slim_globals.cpp
@@ -2337,61 +2337,61 @@ const std::string gSLiM_tsk_individual_metadata_schema_FORMAT_SOURCE = R"V0G0N({
         },
         "tagL0_set": {
 			"index": 10,
-			"type": "bool",
+			"type": "boolean",
 			"binaryFormat": "?",
 			"description": "A flag indicating whether the `tagL0` property is set; if false, accessing `tagL0` is invalid."
         },
         "tagL0": {
 			"index": 11,
-			"type": "bool",
+			"type": "boolean",
 			"binaryFormat": "?",
 			"description": "The `tagL0` property of this individual; only valid if `tagL0_set` is true."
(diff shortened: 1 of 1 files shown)


bhaller:
Huh.  Whoops, sorry!  But looking at the table here:

https://tskit.dev/tskit/docs/stable/metadata.html#numeric-and-boolean-types

how would one know that?  Is there a problem with the doc here?

bhaller:
Merged, thanks!

bhaller:
Hmm, looks like maybe there are further issues with the types in my new schema changes: https://github.com/MesserLab/SLiM/actions/runs/28924838596/job/85809960155.  Apologies; I'm still sick, and on lots of meds, so my brain is quite fuzzy.  But at least in my brain-fuzzed state, it does seem like the doc I linked to above ought to say that "boolean", "integer", and "number" should be used for the "type" property for these various types; that could be another column in the table presented, perhaps.

bhaller:
OK, changed `"type": "float"` to `"type": "number"` for `tagF`, too, so maybe the schema is now correct.  Sorry for the confusion!

petrelharp:
Hm, thanks for fixing the other things? Funny, I built against my changes and then could open a tree sequence; I wonder why I didn't get those errors.

petrelharp:
Re: docs - yes, the docs are very minimal. In their defense, you would have figured this out quickly if you'd tried to use the metadata schema.

Issue #646, opened 2026-07-17
saved mutations can be added later and bypass retained_by_treeseq_


petrelharp:
At least, that's what I think what's going on. First consider this script:
```
initialize() {
    setSeed(23);
	defineConstant("L", 1e2);
	initializeSLiMOptions(keepPedigrees=T, nucleotideBased=T);
    initializeTreeSeq(timeUnit="generations");
	initializeAncestralNucleotides(randomNucleotides(L));
	initializeMutationTypeNuc("m1", 0.5, "f", 0.0);
	initializeGenomicElementType("g1", m1, 1.0, mmJukesCantor(4e-3));
	initializeGenomicElement(g1, 0, L-1);
	initializeRecombinationRate(1e-2);
	defineGlobal("muts", c());
}

1 early() {
	sim.addSubpop("p1", 10);
}

early() {
	defineGlobal("muts", c(muts, sim.mutations));
	catn("Num mutations: " + size(muts));
}

10 late() {
	ind = p1.individuals[0];
	ind.haploidGenome1.addMutations(muts);
   sim.treeSeqOutput("in.trees");
}
```
and then this one:
```
initialize()
{
   setSeed(25);
	defineConstant("L", 1e2);
	initializeSLiMOptions(keepPedigrees=T, nucleotideBased=T);
   initializeTreeSeq(timeUnit="generations");
	initializeMutationTypeNuc("m1", 0.5, "f", 0.0);
   initializeMutationTypeNuc("m2", 0.5, "n", 0.0, 0.1);
	initializeGenomicElementType("g1", m1, 1.0, mmJukesCantor(4e-3));
	initializeAncestralNucleotides(paste0(rep("A", L)));
	initializeGenomicElement(g1, 0, L-1);
	initializeRecombinationRate(1e-2);
}

1 late() { 
    sim.readFromPopulationFile("in.trees");
    sim.treeSeqOutput("out.trees", simplify=F);
}

2 early() {
    catn("Done.");
    sim.simulationFinished();
}
```
which produces 
```
ERROR (Species::__TabulateMutationsFromTables): mutation id 173 referenced by a derived state is not in the mutation metadata table.

Error on script line 16, character 8:

    sim.readFromPopulationFile("in.trees");
```

bhaller:
Fixed at the same time as #648.

PR #647, opened 2026-07-17
small update to implementation of mutation storing

petrelharp:
I've updated this to include a few more details, I think reflecting the current state of things?

In writing this up I wondered about saving/adding mutations, which led to #646, so there might be another note to add.


bhaller:
It's hard to read the diffs on this kind of thing, so I've just assumed that you have done Good Things, and merged.  I'll do another pass over it when #648 has been resolved in one way or another.  :->

Issue #627, opened 2026-05-13
Windows builds, both CLI and GUI, now failing for unclear reason


bhaller:
Here's an example:

https://github.com/MesserLab/SLiM/actions/runs/25653704791

The error, for both builds, looks like:

> Linking CXX executable slim.exe
> D:/a/_temp/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/slim.dir/objects.a(main.cpp.obj):main.cpp:(.text.startup+0x595): undefined reference to `std::basic_filebuf<char, std::char_traits<char> >::_close()'
> collect2.exe: error: ld returned 1 exit status

So it's a link-time error, not a compile-time error.  There are no calls to close() in main.cpp for `slim`, I don't think.  There are two calls to fclose(), but the same two calls to fclose() also occur for main.cpp for `eidostool`, and that built and linked successfully, as can be seen a bit higher in the logs.  So maybe those fclose() calls are just a red herring; or maybe something about the way main.cpp for `eidostool` does things is subtly different, such that the error is not triggered.

Since it emerged spontaneously (I don't think I was working on SLiM at all at the time, it failed in the weekly test build), my guess is that GitHub changed something in their build environment for Windows.  Beyond that I have no idea.  Maybe someone here with Windows knowledge can figure it out?

bhaller:
Tagging some folks who might be able to help: @rdinnager @chris-a-talbot @currocam

currocam:
Hi! I’ve checked the changelog for the direct dependencies setup-msys2 and setup-miniconda but haven’t found anything clear.  I’m also not familiar with Windows.

If you’re willing to use the GitHub action environment as “ground truth”, we can fix the CI by adding `core/main.cpp` to GNULIB_NAMESPACE_SOURCES so the `_close` macro is defined. #629 

https://github.com/currocam/SLiM/blob/4f77dd527b812242a33a38e5e904e99f01009606/windows_compat/gnulib/gllib/unistd.in.h#L383-L386

I’m not sure why adding `main.cpp` wasn’t necessary before, and I haven't fully understood the whole building process in Windows. But it seems to work?

bhaller:
Aha, OK.  I should've tried this, I guess, but the way it appeared spontaneously made me think that this was unlikely to be the problem.  Wrong guess.  :->  Thanks for figuring it out.

[...]

PR #652, opened 2026-07-22
Improve CI (pin actions + Dependabot)

currocam:
Hello Ben, 

I have recently learned a bit more about GitHub Actions working in a project of mine, and I have some ideas about how the current SLiM CI could be improved. 

Before, there have been issues with CI suddenly failing (see #627, #498).

I’m decently sure now that those are triggered by either (1) some minor version being released (say, for example, changing from setup-msys2@v2.31.1 to setup-msys2@v2.32.0 or (2) someone updating a git tag. Note that currently CI sometimes does not pin to any specific version, e.g. actions/cache@v3, conda-incubator/setup-miniconda@v3, etc.

I propose we change the CI to point to specific commit hashes and follow the format `username/repository@commit_hash # tag name`. 

I made that change in the [first commit of this PR](https://github.com/MesserLab/SLiM/commit/cffe57cc6468a3a70bcbf32d77529837078bcdd1). 
### Possible automation

One would have to keep track of those dependencies and update them from time to time. One automation I've found very useful is [dependabot](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/auto-update-actions). It's basically a cron job that opens PRs with the updated dependency. The automated PRs look like this (notice that they maintain the format `commit hash # version tag` and have some useful information in the commit message)

https://github.com/currocam/SLiM/pull/5

Ideally, one would merge those PR after running the updated CI, rather than unexpectedly finding it in red. The obvious downside is that this complicates the CI more. 

### Bump versions

To avoid creating many PRs at once and running CI many times, I’ve squashed all changes from the automated PRs, so the third commit bump all the GitHub Actions to the latest version. Feel free to cherry-pick which commit (if any) you would like to merge. 

Best,
Curro

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    groups:
+      # Batch minor/patch GitHub Actions into a single PR.
+      gha-bump:
+        applies-to: version-updates
+        update-types:
+          - "patch"
+          - "minor"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,13 +11,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Cancel previous runs
-        uses: styfle/cancel-workflow-action@0.12.1
+        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa # 0.12.1
   coverage:
     if: github.event_name != 'schedule' || (github.event_name == 'schedule' && github.repository == 'messerlab/slim')
     runs-on: ubuntu-24.04
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
       - name: Set up GCC
         run: |
           sudo apt-get install gcc-12 g++-12 && \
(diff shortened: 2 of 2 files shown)


bhaller:
Hi @currocam!  This all sounds sensible.

I notice that your dependabot commit addresses only minor and patch updates.  So if your plan that we would manually notice major updates to these tools and do them ourselves manually?  Or am I misunderstanding?  Seems like we probably want to shift forward onto major updates in an automated fashion too, but perhaps not squashed together with the minor/patch updates since we might want to actually think about the major ones (maybe).

Yes, this makes the CI more complex, and no, I don't fully understand it all (and don't really want to :->).  If you are happy to continue keeping things running for the foreseeable future, then I'm happy to adopt the full scheme you propose here.  If not, I might think harder about it, since I don't like being on the hook to maintain things I don't understand.  (But that is true when things break for unknown reasons, too, of course!)

I think modulo your replies to these questions, I'm inclined to merge all of this.  Thanks!

currocam:
(1) The dependabot addresses all updates, I think. What the config file declares is a special group that consists of all minor and patches.  Sorry if the comment was not clear. The idea is that each major update gets its own PR (because presumably it breaks something), whereas all small changes are accumulated into a single PR to reduce the noise.  Also, running the cron job each month sounds sensible to me, but pls adjust. 
(2) Yes, this makes the CI more complex. I’m happy to keep running things :) If at some point it becomes annoying, it should be enough to git rm a single file and close any open PRs.

bhaller:
OK, this all sounds great, thanks!  Merged!  You are a life-saver @currocam.  :->  I'm actively working in the `multitrait` branch these days.  Should I merge these changes into that branch, or is it best for them to remain only on `master`?

currocam:
If it’s configured correctly, the bot should only read the default branch `master`.  So I don’t think there’s any reason not to merge. I would merge to get the latest CI also in the `multitrait`.

PR #659, opened 2026-08-14
fixup of python recipe

petrelharp:
A lot of that wasn't necessary any more!


petrelharp:
(btw, why not have symlinks between the two places recipes are, to avoid having to sync those two sets of files?)

bhaller:
> (btw, why not have symlinks between the two places recipes are, to avoid having to sync those two sets of files?)

Because the trouble involved in copying files over is smaller than the trouble involved in setting up the symlinks and testing that they check in to GitHub properly, that they work on all target platforms (e.g., Windows), etc.?  :->

PR #665, opened 2026-08-22
put binary derived state into the mutation table metadata

### LINKED ISSUE #664: put list of mutation IDs in metadata rather than in derived state
A common thing for people to do is `pyslim.convert_alleles`, which replaces the derived_state with a nucleotide (so that eg we can output to VCF); this doesn't really lose very important information in the previous scheme, but with the new scheme it means you can no longer look up selection coefficients of mutations, etc. @jeffspence noticed that, for instance, stdpopsim does convert_alleles on output by default.

Now we have the option of instead putting the mutation IDs in mutation metadata. This seems preferable for a number of reasons - they'll be in binary, for instance! I've thought of no downsides of this.

The question is: what then goes in derived state? I think the best answer probably is "keep the list of mutation IDs there". This seems a bit odd because then the information will be in two places, but:

- this is somewhat more backwards-compatible
- then alleles will be identical if and only if the alleles actually are identical
- we can explicitly say that although SLiM will output derived states like this, we don't rely on this fact at all, so that for instance SLiM could read back in a tree sequence that had had convert_alleles done on it

bhaller:
This PR fixes #664 (I hope).  Review requested from @petrelharp.  Right now I've tested that it round-trips a tree sequence out to disk and back, and that seems to work AFAICT.  Let me know if it does what you need it to do.

Also, note: it does what it does in an unorthodox way, by munging around inside the tskit data structures, because I didn't want to be making copies of this stuff – the derived state column could potentially be a pretty big buffer, so I didn't want to push the high-water memory usage mark higher if possible.  If there is a cleaner and more orthodox way to do what I'm doing, that would be great of course.  In any case, please make sure that the way that I'm munging tskit's data structures is completely correct and safe (given the current tskit implementation).  This code ought to be bulletproof, of course, but I'm not sure that it is since I don't feel 100% confident in my understanding of how tskit does things under the hood (and AFAIK that is not documented anywhere, even in comments, right?).

### DIFF
--- a/VERSIONS
+++ b/VERSIONS
@@ -208,6 +208,9 @@ multitrait branch:
 	internally, split baselineOffset into a "from substitutions" component and a "from user" component, and expose those components in .trees metadata, to fix #661
 		the top-level JSON schema now has two keys, baselineOffsetFromUser and baselineOffsetFromSubstitutions, that provide these two components
 		SLiM writes both of them, for use on the Python side; when reading a .trees, SLiM uses baselineOffsetFromUser and ignores baselineOffsetFromSubstitutions (calculating that itself)
+	shift to putting binary derived state info in the mutation table metadata, in addition to the ASCII derived state info in its own column, when on disk (i.e., in .trees files); see #664
+		in-memory, the binary derived state info remains in the derived state column and the mutation table metadata is kept empty, as before
+		the schema for the mutation table metadata now encodes an array (of length determined by the metadata itself) of int64_t SLiM mutation IDs
 
 
 version 5.2 (Eidos version 4.2):
--- a/VERSIONS
+++ b/VERSIONS
@@ -209,6 +209,7 @@ multitrait branch:
 		the top-level JSON schema now has two keys, baselineOffsetFromUser and baselineOffsetFromSubstitutions, that provide these two components
 		SLiM writes both of them, for use on the Python side; when reading a .trees, SLiM uses baselineOffsetFromUser and ignores baselineOffsetFromSubstitutions (calculating that itself)
 	shift to putting binary derived state info in the mutation table metadata, in addition to the ASCII derived state info in its own column, when on disk (i.e., in .trees files); see #664
+	    however, we do not rely on the derived state being anything when reading in a .trees file
 		in-memory, the binary derived state info remains in the derived state column and the mutation table metadata is kept empty, as before
 		the schema for the mutation table metadata now encodes an array (of length determined by the metadata itself) of int64_t SLiM mutation IDs
 
--- a/core/slim_globals.cpp
+++ b/core/slim_globals.cpp
@@ -2188,7 +2188,7 @@ const std::string gSLiM_tsk_mutation_metadata_schema_SOURCE = R"V0G0N({
     "additionalProperties": false,
     "codec": "struct",
     "type": "object",
-    "description": "SLiM schema for representing binary derived state data in mutation metadata (the actual mutation metadata is stored in a table in top-level metadata now).",
+    "description": "SLiM schema for representing binary derived state data in mutation metadata (the metadata for each unique SLiM mutation is stored in top-level metadata).",
     "examples": [
         {
             "derived_states": [0, 1, 17]
@@ -2199,7 +2199,7 @@ const std::string gSLiM_tsk_mutation_metadata_schema_SOURCE = R"V0G0N({
             "index": 1,
             "type": "array",
 			"noLengthEncodingExhaustBuffer": true,
-            "description": "An array of SLiM mutation IDs (int64t), representing the (stacked) mutations contained by the derived state for the mutation.  This information is also saved as a comma-separated ASCII string in the derived state column.",
+            "description": "An array of SLiM mutation IDs (int64t), representing the (stacked) mutations contained by the derived state for the mutation.",
             "items": {
                 "binaryFormat": "q",
                 "type": "number"
--- a/core/slim_globals.cpp
+++ b/core/slim_globals.cpp
@@ -2177,7 +2177,39 @@ const std::string gSLiM_tsk_metadata_binary_schema_FORMAT_SOURCE = R"V0G0N({
 
 const std::string gSLiM_tsk_edge_metadata_schema_SOURCE = "";
 const std::string gSLiM_tsk_site_metadata_schema_SOURCE = "";
-const std::string gSLiM_tsk_mutation_metadata_schema_SOURCE = "";		// this is now managed by gSLiM_tsk_metadata_binary_schema in top-level metadata
+
+// actual mutation metadata is now kept in the top-level metadata; see gSLiM_tsk_metadata_binary_schema
+// the mutation metadata column now (on disk only) contains binary derived state info, in addition to the
+// ASCII info kept in the derived state column; see DerivedStatesFromMetadata() and DerivedStatesToMetadata()
+#pragma mark gSLiM_tsk_mutation_metadata_schema_SOURCE
+
+const std::string gSLiM_tsk_mutation_metadata_schema_SOURCE = R"V0G0N({
+    "$schema": "http://json-schema.org/schema#",
+    "additionalProperties": false,
+    "codec": "struct",
+    "type": "object",
+    "description": "SLiM schema for representing binary derived state data in mutation metadata (the actual mutation metadata is stored in a table in top-level metadata now).",
+    "examples": [
+        {
+            "derived_states": [0, 1, 17]
+        }
+    ],
+    "properties": {
+        "derived_states": {
+            "index": 1,
+            "type": "array",
+			"noLengthEncodingExhaustBuffer": true,
+            "description": "An array of SLiM mutation IDs (int64t), representing the (stacked) mutations contained by the derived state for the mutation.  This information is also saved as a comma-separated ASCII string in the derived state column.",
+            "items": {
+                "binaryFormat": "q",
+                "type": "number"
+            }
+        }
+    },
+    "required": [
+        "derived_states"
+    ]
+})V0G0N";
 
 // BCH 12/10/2024: Removed the type field, and changed the treatment of is_vacant.  We have a tricky problem
 // here, which is that is_vacant is now variable-length and there is no count.  The number of byte (uint8_t)
--- a/core/species.cpp
+++ b/core/species.cpp
@@ -8774,70 +8774,102 @@ void Species::CheckAutoSimplification(void)
 void Species::DerivedStatesFromMetadata(tsk_table_collection_t *p_tables)
 {
 	// This is called when reading a .trees file.  On disk, derived state information is kept in the mutation
-	// table's metadata column in binary. When writing out the .trees file we also
-	// put this in the derived state column in ASCII (see #664 for discussion), but we
-	// specifically don't want to assume anything about the derived state column here.
-	// Here we convert back to our in-memory format by swapping the derived state column's ASCII data with the
-	// metadata column's binary data, and then purging the ASCII data from the metadata column.
-	
-	// We want to do this efficiently, without making copies of buffers, etc., so we munge around in tskit's
-	// structs.  Maybe there is a way to do this more cleanly and safely?  :-O
+	// table's metadata column in binary.  When writing out a .trees file we also put this in the derived state
+	// column in ASCII (see #664 for discussion), but we specifically don't want to assume anything about the
+	// derived state column here.  Here we convert back to our in-memory format by swapping the derived state
+	// column's ASCII data with the metadata column's binary data, and then purging the ASCII data from the
+	// metadata column.  To do this efficiently, without making copies of buffers, etc., we use a secondary
+	// mutation table with tsk_mutation_table_takeset_columns(), which is a bit tricky.
 	
 	assert(p_tables != nullptr);
-	assert(p_tables->mutations.derived_state != nullptr);
-	assert(p_tables->mutations.metadata != nullptr);
-	assert(p_tables->mutations.metadata_offset != nullptr);
+	
+	tsk_mutation_table_t &mutation_table = p_tables->mutations;
+	
+	assert(mutation_table.derived_state != nullptr);
+	assert(mutation_table.derived_state_offset != nullptr);
+	assert(mutation_table.metadata != nullptr);
+	assert(mutation_table.metadata_offset != nullptr);
 	
 	// Swap the derived state column and the metadata column; after this, binary derived states will be in the
 	// derived state column and ASCII derived states will be in the metadata column.
-	std::swap(p_tables->mutations.derived_state_length, p_tables->mutations.metadata_length);
-	std::swap(p_tables->mutations.max_derived_state_length, p_tables->mutations.max_metadata_length);
-	std::swap(p_tables->mutations.max_derived_state_length_increment, p_tables->mutations.max_metadata_length_increment);
-	std::swap(p_tables->mutations.derived_state, p_tables->mutations.metadata);
-	std::swap(p_tables->mutations.derived_state_offset, p_tables->mutations.metadata_offset);
+	std::swap(mutation_table.derived_state_length, mutation_table.metadata_length);
+	std::swap(mutation_table.max_derived_state_length, mutation_table.max_metadata_length);
+	std::swap(mutation_table.max_derived_state_length_increment, mutation_table.max_metadata_length_increment);
+	std::swap(mutation_table.derived_state, mutation_table.metadata);
+	std::swap(mutation_table.derived_state_offset, mutation_table.metadata_offset);
 	
 	// Then empty out the metadata column, which we don't use while simulating; we want to free up the buffer
 	// for the metadata, rather than having the (perhaps large capacity) buffer hanging around forever.  The
-	// only way to do that is by freeing the old buffer and mallocing a new minimal buffer.
-	free(p_tables->mutations.metadata);
-	p_tables->mutations.metadata = (char *)malloc(1);	// avoids platform-dependencies on zero-length malloc
-	if (!p_tables->mutations.metadata)
-		EIDOS_TERMINATION << "ERROR (Species::DerivedStatesFromMetadata): allocation failed; you may need to raise the memory limit for SLiM." << EidosTerminate(nullptr);
+	// best way to do this without using (much) private information is to (a) init a new mutation table,
+	// (b) use tsk_mutation_table_takeset_columns() to set up that table with our number of rows but with
+	// empty metadata (setting up whatever internal representation our version of tskit prefers for that),
+	// (c) swap that empty metadata column in to our mutation table, (d) free the metadata column info from the
+	// temporary mutation table (which now contains the old ASCII derived states), and then (e) let that mutation
+	// table go WITHOUT tsk_mutation_table_free().  (It will have pointers to all the same buffers as our own
+	// mutation table, so we don't want to free it!)
+	tsk_mutation_table_t temp_mutation_table;
 	
-	p_tables->mutations.metadata_length = 0;
-	p_tables->mutations.max_metadata_length = 1;
+	tsk_mutation_table_init(&temp_mutation_table, 0);
+	tsk_mutation_table_takeset_columns(&temp_mutation_table,
+									   mutation_table.num_rows,
+									   mutation_table.site,
+									   mutation_table.node,
+									   mutation_table.parent,
+									   mutation_table.time,
+									   mutation_table.derived_state,
+									   mutation_table.derived_state_offset,
+									   /* metadata */ NULL,
+									   /* metadata_offset */ NULL);
 	
-	// Zero out all the metadata offsets; there is an entry for each row, plus one
-	EIDOS_BZERO(p_tables->mutations.metadata_offset, (p_tables->mutations.num_rows + 1) * sizeof(tsk_size_t));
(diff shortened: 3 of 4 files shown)


bhaller:
And again GitHub Desktop has based this PR on `master` instead of `multitrait`.  I _definitely_ told the GUI to base it off of `multitrait`; I was careful!  So, hmm, annoying.  Can you fix it up again, however you do that, @petrelharp?  (Is there UI in GitHub to do it, somewhere?)  [**EDIT:** Found the UI, which is quite hidden!  You have to click the "Edit title" button, and having done that, a popup appears that lets you choose the branch the PR will be merged into.  Their UI designers were smoking something when they came up with that!  :-O]

bhaller:
Hmm.  The branch that this PR is based on, `derived_state_metadata`, still appears to be based on `master`.  So maybe this is still screwed up.  I'm going to just delete this PR, and its associated branch, and try again.  Grr.

bhaller:
OK, clearly I'm (yet again) fundamentally confused by Git's conceptual model.  I just made a new branch, `derived_state_metadata_2`, based on `multitrait`.  I published that branch.  If I now look at it in GitHub, it says "This branch is 335 commits ahead of `master`."  I was expecting it to say "This branch is 0 commits ahead of `multitrait`."  Why is it comparing itself to `master`??  But if I go to Terminal and do `git reflog show derived_state_metadata_2`, it says "Created from multitrait", and it looks like `HEAD` for `derived_state_metadata_2` points to the current head of `multitrait`.  Aargh, so confusing.  So I'm going to hypothesize that the original `derived_state_metadata` branch was OK, and that once I switched this PR over to being based against `master` (why didn't it base off of `multitrait`, since it knows that `derived_state_metadata` is a branch off of `multitrait`??), now this PR is also OK.  So I'm going to reopen this PR, and delete my `derived_state_metadata_2` branch.  But it'd be helpful to understand why Git is acting this way, and to be reassured that merging this PR will not somehow corrupt the project.  Grr, grr.

petrelharp:
The github-ness of this PR looks good - it's based on `multitrait`, and shows 1 commit. Github's display tells you that the *branch* is way ahead of master, because there isn't a way to say that a *branch* is supposed to be compared to a given other branch; it's only a PR that has a notion of "branch to compare to".

core/slim_globals.cpp
```diff
@@ -2177,7 +2177,39 @@ const std::string gSLiM_tsk_metadata_binary_schema_FORMAT_SOURCE = R"V0G0N({
...
+// the mutation metadata column now (on disk only) contains binary derived state info, in addition to the
+// ASCII info kept in the derived state column; see DerivedStatesFromMetadata() and DerivedStatesToMetadata()
+#pragma mark gSLiM_tsk_mutation_metadata_schema_SOURCE
+
+const std::string gSLiM_tsk_mutation_metadata_schema_SOURCE = R"V0G0N({
+    "$schema": "http://json-schema.org/schema#",
+    "additionalProperties": false,
+    "codec": "struct",
+    "type": "object",
+    "description": "SLiM schema for representing binary derived state data in mutation metadata (the actual mutation metadata is stored in a table in top-level metadata now).",
```
petrelharp:
```suggestion
    "description": "SLiM schema for representing binary derived state data in mutation metadata (the metadata for each unique SLiM mutation is stored in top-level metadata).",
```

core/slim_globals.cpp
```diff
@@ -2177,7 +2177,39 @@ const std::string gSLiM_tsk_metadata_binary_schema_FORMAT_SOURCE = R"V0G0N({
...
+        {
+            "derived_states": [0, 1, 17]
+        }
+    ],
+    "properties": {
+        "derived_states": {
+            "index": 1,
+            "type": "array",
+			"noLengthEncodingExhaustBuffer": true,
+            "description": "An array of SLiM mutation IDs (int64t), representing the (stacked) mutations contained by the derived state for the mutation.  This information is also saved as a comma-separated ASCII string in the derived state column.",
```
petrelharp:
```suggestion
            "description": "An array of SLiM mutation IDs (int64t), representing the (stacked) mutations contained by the derived state for the mutation.",
```

core/slim_globals.cpp
```diff
@@ -2177,7 +2177,39 @@ const std::string gSLiM_tsk_metadata_binary_schema_FORMAT_SOURCE = R"V0G0N({
...
+            "items": {
+                "binaryFormat": "q",
```
petrelharp, line 2204: not `Q`? (that'd be [unsigned](https://docs.python.org/3/library/struct.html#format-characters))

VERSIONS
```diff
@@ -208,6 +208,9 @@ multitrait branch:
 	internally, split baselineOffset into a "from substitutions" component and a "from user" component, and expose those components in .trees metadata, to fix #661
 		the top-level JSON schema now has two keys, baselineOffsetFromUser and baselineOffsetFromSubstitutions, that provide these two components
 		SLiM writes both of them, for use on the Python side; when reading a .trees, SLiM uses baselineOffsetFromUser and ignores baselineOffsetFromSubstitutions (calculating that itself)
+	shift to putting binary derived state info in the mutation table metadata, in addition to the ASCII derived state info in its own column, when on disk (i.e., in .trees files); see #664
+		in-memory, the binary derived state info remains in the derived state column and the mutation table metadata is kept empty, as before
+		the schema for the mutation table metadata now encodes an array (of length determined by the metadata itself) of int64_t SLiM mutation IDs
```
petrelharp, line 214:
```suggestion
	shift to putting binary derived state info in the mutation table metadata, in addition to the ASCII derived state info in its own column, when on disk (i.e., in .trees files); see #664
	    however, we do not rely on the derived state being anything when reading in a .trees file
		in-memory, the binary derived state info remains in the derived state column and the mutation table metadata is kept empty, as before
		the schema for the mutation table metadata now encodes an array (of length determined by the metadata itself) of int64_t SLiM mutation IDs
```

petrelharp (commented):
I haven't got into the weeds on the main work here yet because I think there's a much simpler way to do this. The `derived_state` and `metadata` columns are stored in exactly the same way by tskit. So, why not just move around the pointers instead of looping over entries? So we want to move the pointers `metadata` and `metadata_length` over to `derived_state` and `derived_state_length`, and then set `metadata` to `NULL` and `metadata_length` to 0?

(Tangentially, this made me think "why not just go through the code and (more or less) change over the references to `derived_state` to `metadata`?" But this is no good because we make heavy use of the `tsk_variant` object, which is very useful and relies on ancestral/derived states.)

petrelharp:
Ah also I think we want "we don't actually use `derived_state` for anything, but make it an ASCII copy of the comma-separated list of IDs" as a policy, which isn't reflected in the comments here, but should be? (We should also update `treerec/implementation.md`?)

core/species.cpp
```diff
@@ -8771,115 +8771,101 @@ void Species::CheckAutoSimplification(void)
...
-		if (ret < 0) handle_error("derived_from_ascii", ret);
-	}
+	// We want to do this efficiently, without making copies of buffers, etc., so we munge around in tskit's
+	// structs.  Maybe there is a way to do this more cleanly and safely?  :-O
+	
+	assert(p_tables != nullptr);
+	assert(p_tables->mutations.derived_state != nullptr);
+	assert(p_tables->mutations.metadata != nullptr);
+	assert(p_tables->mutations.metadata_offset != nullptr);
+
```
petrelharp, line 8801: Like here can't we just do (edited):
```suggestion
p_tables->mutations.derived_state = p_tables.mutations.metadata;
p_tables->mutations.derived_state_length = p_tables.mutations.metadata_length;
p_tables->mutations.metadata = nullptr;
p_tables->mutations.metadata_length = zeros; // <-- the easy way in C++ to do a vector of zeros
```

bhaller:
> I haven't got into the weeds on the main work here yet because I think there's a much simpler way to do this. The `derived_state` and `metadata` columns are stored in exactly the same way by tskit. So, why not just move around the pointers instead of looping over entries? So we want to move the pointers `metadata` and `metadata_length` over to `derived_state` and `derived_state_length`, and then set `metadata` to `NULL` and `metadata_length` to 0?

That's pretty much what the code does.  That's the purpose of the `std::swap()` calls.  But I can't leave the the metadata pointer as `NULL`, I think?  All the columns are supposed to be allocated, aren't they?  That was my understanding, but like I said, I am not aware of ANY documentation on how the policies these internal data structures are supposed to follow.  It's all just in Jerome and Ben's head (and yours, if you understand it all).  It drives me crazy, their total opposition to commenting code.  Anyhow, if you look at the code for `Species::DerivedStatesFromMetadata()` it does exactly what you suggest, just with a 1-byte allocation rather than `NULL`.  `Species::DerivedStatesToMetadata()` is more complicated because it still has to convert the binary derived state info into ASCII.

> (Tangentially, this made me think "why not just go through the code and (more or less) change over the references to `derived_state` to `metadata`?" But this is no good because we make heavy use of the `tsk_variant` object, which is very useful and relies on ancestral/derived states.)

Yes, the derived state column needs to be functional, I think?  Maybe for simplify as well?

bhaller:
> ...there isn't a way to say that a _branch_ is supposed to be compared to a given other branch; it's only a PR that has a notion of "branch to compare to"...

Huh, how strange.  When you make a branch, you make it based on another branch; and `reflog` knows that it is based on that other branch; and if you make a PR from it, you will of course want that PR to be compared to that other branch (at least as a default); so why not show the comparison to that other branch, then?  Git knows it...?  This is a very strange and confusing design decision.

petrelharp:
> That's pretty much what the code does. That's the purpose of the std::swap() calls. 

Ah, good.

> But I can't leave the the metadata pointer as NULL, I think? All the columns are supposed to be allocated, aren't they?

Oh, you're probably right.

> That was my understanding, but like I said, I am not aware of ANY documentation on how the policies these internal data structures are supposed to follow.

Have you read the [C API docs](https://tskit.dev/tskit/docs/latest/c-api.html)?  It's kinda minimal but it is SOME documentation.

core/species.cpp
```diff
@@ -8771,115 +8771,101 @@ void Species::CheckAutoSimplification(void)
...
 
-void Species::DerivedStatesFromAscii(tsk_table_collection_t *p_tables)
+void Species::DerivedStatesFromMetadata(tsk_table_collection_t *p_tables)
 {
-	// This modifies p_tables in place, replacing the derived_state column of p_tables with a binary version.
-	tsk_mutation_table_t mutations_copy;
-	int ret = tsk_mutation_table_copy(&p_tables->mutations, &mutations_copy, 0);
-	if (ret < 0) handle_error("derived_from_ascii", ret);
+	// This is called when reading a .trees file.  On disk, derived state information is kept in the mutation
+	// table's metadata column in binary, AND in the derived state column in ASCII (see #664 for discussion).
```
petrelharp:
```suggestion
	// table's metadata column in binary. When writing out the .trees file we also
	// put this in the derived state column in ASCII (see #664 for discussion), but we
	// specifically don't want to assume anything about the derived state column here.
```

petrelharp:
Also: I think the reason that the sort of documentation you're looking for here doesn't exist is because we haven't settled on a public and stable API for it. We want to have the freedom to change how some of this works under the hood.

bhaller:
> Have you read the [C API docs](https://tskit.dev/tskit/docs/latest/c-api.html)? It's kinda minimal but it is SOME documentation.

I've skimmed it and found it pretty useless for my purposes.  For example, on the salient question of whether you're allowed to free a column and put `NULL` there, with a zero length, I don't see that it says anything at all.  Or, when I put a 1-byte malloced pointer there instead, since I was afraid `NULL` isn't allowed, so I do:

```
	free(p_tables->mutations.metadata);
	p_tables->mutations.metadata = (char *)malloc(1);	// avoids platform-dependencies on zero-length malloc
	...
	p_tables->mutations.metadata_length = 0;
	p_tables->mutations.max_metadata_length = 1;
```

Is _that_ allowed?  I have no idea.  What are the rules and requirements for how columns are managed?  I don't see any guidance on that at all.  (If I have missed it somewhere, then of course I'll eat my words; but I've mentioned this before and have gotten nowhere.)

bhaller:
> Also: I think the reason that the sort of documentation you're looking for here doesn't exist is because we haven't settled on a public and stable API for it. We want to have the freedom to change how some of this works under the hood.

I don't think that's a valid reason not to document code.  You can document it and write "This is internal and subject to change."  I've had that debate with Jerome and Ben.  :->

core/species.cpp
```diff
@@ -8771,115 +8771,101 @@ void Species::CheckAutoSimplification(void)
...
-										 mutations_copy.parent,
-										 mutations_copy.time,
-										 (char *)binary_derived_state.data(),
-										 binary_derived_state_offset.data(),
-										 mutations_copy.metadata,
-										 mutations_copy.metadata_offset);
-		if (ret < 0) handle_error("derived_from_ascii", ret);
-	}
+	// We want to do this efficiently, without making copies of buffers, etc., so we munge around in tskit's
+	// structs.  Maybe there is a way to do this more cleanly and safely?  :-O
```
petrelharp: A cleaner and safer way to do all this would be to use [`tsk_mutation_table_set_columns`](https://tskit.dev/tskit/docs/latest/c-api.html#c.tsk_mutation_table_set_columns) (passing in `metadata` to `derived_state`, and leaving `metadata` NULL).

I think we should probably do this, even if it makes a copy. But does it? Let's see: `set_columns` [just calls `clear`](https://github.com/tskit-dev/tskit/blob/main/c/tskit/tables.c#L4344) and then [`append_columns`](https://github.com/tskit-dev/tskit/blob/dcef7cac257a25288d7e2dee6f35483bd3ddc374/c/tskit/tables.c#L4344); this then does a [`memcpy`](https://github.com/tskit-dev/tskit/blob/dcef7cac257a25288d7e2dee6f35483bd3ddc374/c/tskit/tables.c#L4367) so it looks like `set_columns` does make a copy.

Hm, but how about [`tsk_mutation_table_takeset_columns`](https://github.com/tskit-dev/tskit/blob/dcef7cac257a25288d7e2dee6f35483bd3ddc374/c/tskit/tables.h#L4661)? This is public (in `tables.h`), so fair game for use (the lack of docstring just means that there isn't a no-backwards-incompatibility guarantee).  Hm, yes! That appears to be doing just what we want!

petrelharp: The swapping you're doing seems fine; it's the "empty out the metadata column" which seems more suspicious.

petrelharp: So since `takeset_columns` does a `free` to `self` first, the way to do this is to (a) init a new mutation table; (b) takeset the columns to it; (c) set the columns on the old one to NULL; (d) free the old one; (e) change `p_tables->mutations` to the new one?

Why `free` the old one? It looks like the only thing that [`tsk_mutation_table_free`](https://github.com/tskit-dev/tskit/blob/dcef7cac257a25288d7e2dee6f35483bd3ddc374/c/tskit/tables.c#L4091) does is (1) free the columns (unncessary because we've taken over their memory) and (2) freeing the metadata schema. So if we didn't free the old one we'd have a (small) memory leak: the metadata schema. So alternatively instead of (c) and (d), just free the metadata schema on the old one. This requires a comment saying that if anything else gets added to the `mutation_table_t` that requires free'ing then it'd need changing.

core/species.cpp
```diff
@@ -8771,115 +8771,101 @@ void Species::CheckAutoSimplification(void)
...
-	tsk_mutation_table_t mutations_copy;
-	int ret = tsk_mutation_table_copy(&p_tables->mutations, &mutations_copy, 0);
-	if (ret < 0) handle_error("derived_to_ascii", ret);
+	// This is called when saving out a .trees file.  On disk, derived state information is kept in the mutation
+	// table's metadata column in binary, AND in the derived state column in ASCII (see #664 for discussion).
+	// Here we convert from our in-memory format by swapping the derived state column's binary data with the
+	// empty metadata column, and then adding ASCII derived state information.
+	
+	// We want to do this efficiently, without making copies of buffers, etc., so we munge around in tskit's
+	// structs.  Maybe there is a way to do this more cleanly and safely?  :-O
```
petrelharp: This seems fine, but alternatively it seems like you could construct the `derived_state` and `derived_state_offset` columns as below, then do `takeset_columns` just as above to stick them in (and move the previous derived state to metadata)?

core/slim_globals.cpp
bhaller: Yeah, not Q.  I dislike unsigned types in general; when you truly need the extra bit of storage space for the range of values you expect, that's justifiable, but in general I think unsigned types are bug-prone and are not a good idea.  `slim_mutationid_t` is `int64_t`.

core/species.cpp
bhaller: OK, stay tuned for my next commit.  :->

core/species.cpp
bhaller: Yep, see next commit (assuming this idea pans out :->).

bhaller:
OK, accepted/replied/resolved review stuff.  I'll now work on improving the code with `tsk_mutation_table_takeset_columns()` as suggested.  :->

bhaller:
OK, updated with `tsk_mutation_table_takeset_columns()` as discussed.  (Although I followed a slightly different strategy than you proposed, to avoid needing to malloc the temporary mutation table or swap it in.)  A final review would be great, and then hopefully this is ready to merge.  Thanks!

core/species.cpp
```diff
@@ -8771,115 +8771,160 @@ void Species::CheckAutoSimplification(void)
...
+	
+	tsk_mutation_table_init(&temp_mutation_table, 0);
+	tsk_mutation_table_takeset_columns(&temp_mutation_table,
+									   mutation_table.num_rows,
+									   mutation_table.site,
+									   mutation_table.node,
+									   mutation_table.parent,
+									   mutation_table.time,
+									   mutation_table.derived_state,
+									   mutation_table.derived_state_offset,
```
petrelharp, line 8829: Why do you need those `swap`s above? Why not just
```suggestion
									   mutation_table.metadata,
									   mutation_table.metadata_offset,
```

petrelharp: oh wait reading the comment now

petrelharp: Wait, no, I don't get it - why is all this swapping necessary? Why can't you do:

1. takeset to the temporary table, with the `metadata` in the `derived_state` slot
2. set `p_tables->mutations` to the (pointer to the) temporary table
3. free the original table's metadata schema?

core/species.cpp
```diff
@@ -8771,115 +8771,160 @@ void Species::CheckAutoSimplification(void)
...
+									   /* metadata */ new_metadata_buffer,
+									   /* metadata_offset */ text_derived_state_offset);
+
```
petrelharp, line 8926: okay, and this is more confusing here: why swap -> free schema -> swap back? what are the swaps doing?

core/species.cpp
bhaller: I don't like changing `p_tables->mutations` to the temporary table, for a couple of reasons.  (1) It requires that the temporary table be malloced, so it adds an allocation call (which is slow) and a free.  (2) Changing over to a new pointer adds a risk of bugs where someone has a pointer to the table and now their pointer is invalid.  So I prefer the way I've written it.  I want the temporary table to remain the temporary table, and to fix up the original table to be the way it ought to be.

core/species.cpp
bhaller: The second set of swaps is not a "swap back"; it is swapping the derived state and metadata columns within the single mutation table.  That is necessary to put things where they belong, since the binary derived state information was being kept in the derived state column, and now needs to be in the metadata column (and vice versa).  I think the code is correct and necessary as it stands (given that I don't want to change the pointer to the mutation table in the table collection); have another look.

petrelharp:
BTW: are you sure that [`std::swap`](https://en.cppreference.com/cpp/algorithm/swap) actually doesn't copy the contents? I see that maybe [`std::vector::swap`](https://stackoverflow.com/questions/65237774/what-does-stdvectorswap-actually-do) doesn't, but is this a "vector"? (Not that this is important that we don't copy, btw.)

core/species.cpp
petrelharp: Hah, sorry yes.

bhaller:
> BTW: are you sure that [`std::swap`](https://en.cppreference.com/cpp/algorithm/swap) actually doesn't copy the contents? I see that maybe [`std::vector::swap`](https://stackoverflow.com/questions/65237774/what-does-stdvectorswap-actually-do) doesn't, but is this a "vector"? (Not that this is important that we don't copy, btw.)

Yep, I'm quite sure of that.

core/species.cpp
petrelharp: FYI, and probably you know this, but the current code does a `malloc`, since `takeset` calls [`takeset_ragged_column`](https://github.com/tskit-dev/tskit/blob/b95d2262aac9d7c63ef6e174c1608f50995918ee/c/tskit/tables.c#L687) which calls [`alloc_empty_ragged_column`](https://github.com/tskit-dev/tskit/blob/main/c/tskit/tables.c#L658) for the empty columns.

bhaller:
Merging after code review and Slack discussion.

core/species.cpp
bhaller: Yeah; no way I can prevent that, with tskit's APIs.  (It surprises me that they chose to make empty columns still be malloced rather than nullptr; that's not the choice I would've made.)

PR #668, opened 2026-08-26
shift the migrant flag from metadata to the flags column

### LINKED ISSUE #666: remove `flags` from individual metadata?
Currently we use both the `individual.flags` property (to record "remembered", "retained", and "alive") and a special "flags" property in metadata (to record 'is a migrant"). This is because when we first defined metadata, the thinking was that `individual.flags` should be kept for tskit-specific things. But then later the thinking shifted to "client code should use higher-order bits" (thus, we use bits 16, 17, and 18).

Also, turns out that `pyslim.INDIVIDUAL_FLAG_MIGRATED` was wrong (it was "1 << 1", which is 2, not 1, which is what it's supposed to be). So I don't think anyone has ever used the flags entry in metadata (or if they did, I'll bet they got wrong answers).

So if we're breaking things, then I propose removing the `flags` property from individual metadata, for efficiency and simplicity. We can put "migrated" in the `individual.flags`, as `2**19`.

bhaller:
fix #666, shift the migrant flag from metadata to the flags column


core/species.cpp
```diff
@@ -11064,34 +11081,33 @@ void Species::MetadataForIndividual(Individual *p_individual, IndividualMetadata
...
-	p_metadata->tag_ = p_individual->tag_value_;
-	p_metadata->tagF_ = p_individual->tagF_value_;
+	p_metadata->sex_ = (int16_t)p_individual->sex_;		// IndividualSex, but int16_t in the record
 	p_metadata->tagL0_set_ = p_individual->tagL0_set_;
 	p_metadata->tagL0_ = p_individual->tagL0_set_ ? p_individual->tagL0_value_ : false;
 	p_metadata->tagL1_set_ = p_individual->tagL1_set_;
 	p_metadata->tagL1_ = p_individual->tagL1_set_ ? p_individual->tagL1_value_ : false;
 	p_metadata->tagL2_set_ = p_individual->tagL2_set_;
 	p_metadata->tagL2_ = p_individual->tagL2_set_ ? p_individual->tagL2_value_ : false;
+
```
petrelharp, line 11110:
```suggestion
```

core/species.cpp
```diff
@@ -11993,6 +12010,13 @@ void Species::__TabulateSubpopulationsFromTreeSequence(std::unordered_map<slim_o
 		
 		ts_subpop_info &subpop_info = subpop_info_iter->second;
 		
+		// remember the SLIM_TSK_INDIVIDUAL_MIGRATED flag from the flags column of the individual table
+		// this used to be in metadata, but now that it is in the flags column we need to keep it separately
+		if (individual.flags & SLIM_TSK_INDIVIDUAL_MIGRATED)
+			subpop_info.migrant_flags_.push_back(true);
+		else
+			subpop_info.migrant_flags_.push_back(false);
```
petrelharp:
```suggestion
		subpop_info.migrant_flags_.push_back(individual.flags & SLIM_TSK_INDIVIDUAL_MIGRATED);
```
?

core/species.cpp
petrelharp: oh I see is this linebreak indicating byte alignment or something?

petrelharp (commented):
LGTM! I'll have a look at the python tests next.

petrelharp:
So far I'm pretty sure the test errors are because the individual metadata isn't getting written out correctly. Trying to figure out how to verify that.

core/species.h
```diff
@@ -145,37 +145,37 @@ typedef struct __attribute__((__packed__)) {
...
-	slim_objectid_t subpopulation_id_;      // 4 bytes (int32_t): the subpopulation the individual belongs to
-	int32_t sex_;							// 4 bytes (int32_t): the sex of the individual, as defined by the IndividualSex enum
-	uint32_t flags_;						// 4 bytes (uint32_t): assorted flags, see below
 	int64_t tag_;							// 8 bytes (int64_t): the `tag` property value
 	double tagF_;							// 8 bytes (double): the `tagF` property value
+	
+	slim_age_t age_;                        // 4 bytes (int32_t): the age of the individual (-1 for WF models)
+	slim_objectid_t subpopulation_id_;      // 4 bytes (int32_t): the subpopulation the individual belongs to
+	
+	int16_t sex_;							// 2 bytes (int16_t): the sex of the individual, as defined by the IndividualSex enum
```
petrelharp, line 154: here the number of bytes changed, but it didn't change [in the schema](https://github.com/MesserLab/SLiM/pull/668/changes#diff-a51b99d77dc75cfce8aeac5c26d419253644c73cde5392ea86ebc848b55173bdR2358)

core/slim_globals.cpp
```diff
@@ -2335,104 +2328,98 @@ const std::string gSLiM_tsk_individual_metadata_schema_FORMAT_SOURCE = R"V0G0N({
...
+            "index": 7,
             "type": "integer",
             "binaryFormat": "i",
             "description": "The ID of the subpopulation the individual was part of, either when the tree sequence was written out (if the individual was alive then), or the last time they were Remembered (if not)."
         },
         "sex": {
-            "index": 6,
+            "index": 8,
             "type": "integer",
             "binaryFormat": "i",
```
petrelharp: should be `"h"`?

core/species.cpp
bhaller: I don't think this would work, although I'm not sure.  `individual.flags & SLIM_TSK_INDIVIDUAL_MIGRATED` will be either 0 to 1 << 19, whereas `subpop_info.migrant_flags_` is a vector of `bool` and is thus expecting to get a 0 or 1.  But I can clean it up a bit, the if-else is a bit messy.  :->

core/species.cpp
bhaller: Yeah, exactly.  My own bookkeeping mnemonic.  Matches the linebreaks in the header declaration of the metadata struct.

bhaller:
OK, good to go!  Thanks for the review, and for finding that bug!

PR #669, opened 2026-08-26
error in schema and test machinery

petrelharp:
Whoopsie!


bhaller:
Good stuff, thanks for patching all this up!

Issue #676, opened 2026-09-13
clarify derived state / metadata[derived states] distinction for mutations


petrelharp:
Thanks to #665, now the SLiM IDs of mutations carried by a tskit mutation are recorded in two places: the mutation's `derived_state`, and in it's `metdata["derived_states"]`. Ideally, when reading things back in, we would ignore the former and only use the latter, but this is complicated because we're using the `tsk_variant` machinery, which returns alleles. So to use the metadata instead we'd need to either extend the `tsk_variant` behavior or write our own. This would be fine but also not top priority.

So currently, to read in the file we need the `derived_state` column to have the comma-separated list of mutations IDs in it. Are there any requirements on the metadata? Does the code check that these agree? We should record this in `treerec/implementation.md` and in the manual if it's not there already (I'm only partway through checking).

bhaller:
The metadata is now documented as matching the derived state, but in binary rather than ASCII.  Yes, this should be in implementation.md and should be checked by a python test; I can do the former, you should do the latter.  :->  Sound good?

bhaller:
OK, I just pushed a first pass at the `implementation.md` doc for this.  It is already documented in the manual, I think; should be in the latest draft that I posted in stdpopsim.  :->

petrelharp:
From the docs:
<img width="881" height="839" alt="Image" src="https://github.com/user-attachments/assets/65cf9f61-c7ed-4ef2-90f9-94c603fd7543" />
I think this should say something like:
> When SLiM reads in a tree sequence, it uses the derived state (*not* the metadata) to assign mutations to haplosomes.

and then perhaps,
> It does not at present do anything with the list of derived states in mutation metadata (or even check for agreement between the list of mutation IDs in derived state and in metadata).

or the analogous statement if in fact it *does* do that consistency check?
Finally,
> However, this behavior may change in the future, so both derived state and metadata should be present.

petrelharp:
(This is kinda making me want to change the code so it uses metadata and not the derived state.)

petrelharp:
The note in `implementation.md` in [the recent commit](https://github.com/MesserLab/SLiM/commit/84cf290f84eb031306772e030665d0d1456cf61d) is good (although I think a bit redundant with elsewhere in the file?) but it's missing the key note about what is the source of truth? (ie there's redundancy; so which is the source of truth and is agreement between them required?)

bhaller:
> I think this should say something like:
> 
[...]

PR #680, opened 2026-09-15
Zero ancestral state when reading in tree sequence

petrelharp:
As discussed in #676?

Except this doesn't currently work; not sure why atm.


petrelharp:
Got it; however, I also got a simpler way. The following does nearly the same thing:
```
	tsk_safe_free(site_table.ancestral_state);
	tsk_safe_free(site_table.ancestral_state_offset);
	site_table.ancestral_state = zero_ancestral_state;
	site_table.ancestral_state_offset = zero_ancestral_state_offset;
	site_table.ancestral_state_length = 0;
	site_table.max_ancestral_state_length = 0;
	site_table.max_ancestral_state_length_increment = 0;
```
The last two things are appropriate because the requirement on `max_length` is that it be *no greater than* "how much memory have we allocated and so can fill up before we have to alloc more here" and the `increment` can be set to 0 (in which case it doubles by default); see in tskit `calculate_max_length` and `expand_ragged_column`.

I'm also not totally sure that the longer and more complicated way is actually more safe? In understanding how this works, I'm confident that either way works, so up to you. I could (now) rewrite the mutation table code in this simpler way if you like.

petrelharp:
Another stochastic failure, looks like?
Screenshot From 2026-09-15 10-58-49

bhaller:
> Got it; however, I also got a simpler way. The following does nearly the same thing:
> 
> ```
> 	tsk_safe_free(site_table.ancestral_state);
> 	tsk_safe_free(site_table.ancestral_state_offset);
> 	site_table.ancestral_state = zero_ancestral_state;
> 	site_table.ancestral_state_offset = zero_ancestral_state_offset;
> 	site_table.ancestral_state_length = 0;
> 	site_table.max_ancestral_state_length = 0;
> 	site_table.max_ancestral_state_length_increment = 0;
> ```
> 
> The last two things are appropriate because the requirement on `max_length` is that it be _no greater than_ "how much memory have we allocated and so can fill up before we have to alloc more here" and the `increment` can be set to 0 (in which case it doubles by default); see in tskit `calculate_max_length` and `expand_ragged_column`.
> 
> I'm also not totally sure that the longer and more complicated way is actually more safe? In understanding how this works, I'm confident that either way works, so up to you. I could (now) rewrite the mutation table code in this simpler way if you like.

Groovy.  I prefer what's in the diffs.  The problem with your alternative code is that, yes, it works at the moment, but it relies on internal behavior of tskit that is not documented and is therefore free to change in a later version.  The version in the diffs is more robust to changes that could happen in the future, I think.  (If it is ever decided, in the tskit group, that these kinds of aspects of the internal implementation are not going to change, it would be great if that could be made official, and this stuff could actually be documented.)

So.  Ready to merge?

bhaller:
I think this is ready to merge, and I have cell signal now, so I'm gonna do it.  :->

